PR #71, opened 2019-05-23
Update eslint-plugin-node peerDependency to 9

spearl:
eslint-plugin-node is at 9.1.0

We should upgrade this peerDependency to ^9.0.0 or at least >=8.0.0


mightyiam:
Thank you.

PR #96, opened 2019-08-05
feat: update ESLint & TypeScript plugin

mightyiam:
BREAKING CHANGE: `eslint-config-standard` has been updated to `13.0.1`.
BREAKING CHANGE: Updatd all peer dependencies. Fixes #80. Fixes #94.

Added option for rule `@typescript-eslint/explicit-function-return-type`: `allowHigherOrderFunctions: true`.

BREAKING CHANGE: `typescript-eslint/parser` is no longer a `peerDependency`, but a `dependency`. Users should remove it from their configuration, as well. Closes #65.

Updated `@typescript-eslint/eslint-plugin` to `1.10.2`.

PR #105, opened 2019-08-05
feat: update ESLint & TypeScript plugin

### LINKED ISSUE #65: Enjoy the benefits of ESLint's new simplified config
Read the documentation changes in this PR to understand: https://github.com/eslint/eslint/pull/11388/files#diff-33bbe98f532eb757e2789f1a3549c9ca

### LINKED ISSUE #80: Dependencies different from standard
Thanks for this package. Hopefully, they will merge these rules.

Why are the packages required to run this listed as peer dependencies instead of normal dependencies? `standard` just has them listed like this: [standard/package.json#L14-L22](https://github.com/standard/standard/blob/155d52710f587439d58b146f790df322336e195a/package.json#L14-L22)

Also is there a reason that the version requirements are different to the ones standard are using or is just that you're using greenkeeper? It has caused some problems running regression tests against node v6. Namely using version 9 of `eslint-plugin-node` instead of 7 breaks compatibility with node 6.

### LINKED ISSUE #94: peerDependencies should be superset of those of eslint-config-standard

mightyiam:
BREAKING CHANGE: `eslint-config-standard` has been updated to `13.0.1`.
BREAKING CHANGE: Updatd all peer dependencies. Fixes #80. Fixes #94.

Added option for rule `@typescript-eslint/explicit-function-return-type`: `allowHigherOrderFunctions: true`.

BREAKING CHANGE: `typescript-eslint/parser` is no longer a `peerDependency`, but a `dependency`. Users should remove it from their configuration, as well. Closes #65.

Updated `@typescript-eslint/eslint-plugin` to `1.10.2`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,12 +1,13 @@
 export = {
+  parser: '@typescript-eslint/parser',
   extends: 'eslint-config-standard',
   plugins: ['@typescript-eslint'],
   overrides: [
     {
       files: ['*.ts', '*.tsx'],
       rules: {
-        'camelcase': 'off', // in favor of TypeScript rule
-        'indent': 'off',
+        camelcase: 'off', // in favor of TypeScript rule
+        indent: 'off',
         'no-array-constructor': 'off',
         'no-undef': 'off', // TypeScript has this functionality by default
         'no-unused-vars': 'off', // in favor of TypeScript rule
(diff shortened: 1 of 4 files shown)


mightyiam:
@feross, this is identical to #96, with the difference that it is rebased on `master`, getting rid of the reason for a failing CI run: an un-conventional commit message.

I would have rebased #96, but org contributing guidelines ask to not edit history.

src/index.ts
```diff
@@ -1,34 +1,37 @@
...
   extends: 'eslint-config-standard',
   plugins: ['@typescript-eslint'],
   overrides: [
     {
       files: ['*.ts', '*.tsx'],
       rules: {
-        'camelcase': 'off', // in favor of TypeScript rule
-        'indent': 'off',
+        camelcase: 'off', // in favor of TypeScript rule
+        indent: 'off',
```
LinusU: I think that the new rule in Standard 14 is that all lines should have `'` if at least one row needs it 😄 🤔

feross: @LinusU Actually, we went with the rule that we only need `'` on lines where it's needed. In other words, this is okay:

```js
const obj = {
  'hi-there': 1,
  hey: 2
}
```

PR #115, opened 2019-08-08
test: add a deepEqual exports test

mightyiam:


src/index.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          '@typescript-eslint/promise-function-async': 'error',
+          '@typescript-eslint/restrict-plus-operands': 'error',
+          '@typescript-eslint/require-array-sort-compare': 'error',
+          '@typescript-eslint/type-annotation-spacing': 'error'
+        }
+      }
+    ]
+  }
+
+  t.deepEqual(exported, expected)
```
mightyiam: This should help make sure that changes that rearrange properties don't make more changes than they intend to.

LinusU:
I personally think that this seems a bit strange 🤔

We are basically defining the rules in two files and asserting that they are the same, meaning we have to update in two places whenever we make changes...

Any opinions from other maintainers? ☺️

swansontec:
The motivation is #118, which re-orders the rules and adds comments. I think @mightyiam wants to verify that that particular pull request doesn't change anything of substance, by comparing the before & after cases for equality.

This is very similar to the what Jest shapshots do, but in a less-automated way. Using Jest instead of Ava for this could remove the need for the code duplication (although the snapshots would still be stored in git).

feross:
This personally seems like overkill to me. How often do you re-order the list? Can't we just alphabetize it once and then refuse all future PRs from untrusted users who just re-arrange or re-order things?

swansontec:
Since I committed #118 on top of this change, a maintainer could just check out that branch, run `yarn && yarn test` once to verify that #118 doesn't change anything, and then cherry-pick #118 onto master.

mightyiam:
Yes, what @swansontec explained is correct. It makes me feel comfortable and I am willing to support it. This is so small, that updating in two places doesn't seem like considerable extra work, to me. Just a few seconds more. It might seem weird but it makes me happy.

mightyiam:
I'll merge this in order to continue and if it bothers you and you'd like it removed after this reordering, then PR, I guess.

PR #134, opened 2019-08-22
feat: @typescript-eslint/eslint-plugin@^1.12.0

mightyiam:
BREAKING CHANGE: added rule `@typescript-eslint/prefer-readonly` https://github.com/typescript-eslint/typescript-eslint/commit/76b89a5
BREAKING CHANGE: added rule `@typescript-eslint/strict-boolean-expressions` https://github.com/typescript-eslint/typescript-eslint/commit/34e7d1e
BREAKING CHANGE: replaced rule `@typescript-eslint/no-triple-slash-reference` with `@typescript-eslint/no-reference-import` https://github.com/typescript-eslint/typescript-eslint/commit/af70a59

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -60,16 +60,18 @@ test('export', (t): void => {
           '@typescript-eslint/no-non-null-assertion': 'error',
           '@typescript-eslint/no-object-literal-type-assertion': 'error',
           '@typescript-eslint/no-this-alias': ['error', { allowDestructuring: true }],
-          '@typescript-eslint/no-triple-slash-reference': 'error',
           '@typescript-eslint/no-unnecessary-type-assertion': 'error',
           '@typescript-eslint/no-unused-vars': ['error', { args: 'none' }],
           '@typescript-eslint/no-use-before-define': ['error', { functions: false, classes: false, variables: false, typedefs: false }],
           '@typescript-eslint/no-useless-constructor': 'error',
           '@typescript-eslint/no-var-requires': 'error',
           '@typescript-eslint/prefer-function-type': 'error',
+          '@typescript-eslint/prefer-readonly': 'error',
           '@typescript-eslint/promise-function-async': 'error',
           '@typescript-eslint/restrict-plus-operands': 'error',
           '@typescript-eslint/require-array-sort-compare': 'error',
+          '@typescript-eslint/strict-boolean-expressions': 'error',
+          '@typescript-eslint/triple-slash-reference': ['error', { lib: 'never', path: 'never', types: 'never' }],
           '@typescript-eslint/type-annotation-spacing': 'error'
         }
       }
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,13 +66,15 @@ export = {
         '@typescript-eslint/no-non-null-assertion': 'error',
         '@typescript-eslint/no-object-literal-type-assertion': 'error',
         '@typescript-eslint/no-this-alias': ['error', { allowDestructuring: true }],
-        '@typescript-eslint/no-triple-slash-reference': 'error',
         '@typescript-eslint/no-unnecessary-type-assertion': 'error',
         '@typescript-eslint/no-var-requires': 'error',
         '@typescript-eslint/prefer-function-type': 'error',
+        '@typescript-eslint/prefer-readonly': 'error',
         '@typescript-eslint/promise-function-async': 'error',
         '@typescript-eslint/require-array-sort-compare': 'error',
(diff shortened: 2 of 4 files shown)


src/index.test.ts
```diff
@@ -60,16 +60,18 @@ test('export', (t): void => {
...
           '@typescript-eslint/no-object-literal-type-assertion': 'error',
           '@typescript-eslint/no-this-alias': ['error', { allowDestructuring: true }],
-          '@typescript-eslint/no-triple-slash-reference': 'error',
           '@typescript-eslint/no-unnecessary-type-assertion': 'error',
           '@typescript-eslint/no-unused-vars': ['error', { args: 'none' }],
           '@typescript-eslint/no-use-before-define': ['error', { functions: false, classes: false, variables: false, typedefs: false }],
           '@typescript-eslint/no-useless-constructor': 'error',
           '@typescript-eslint/no-var-requires': 'error',
           '@typescript-eslint/prefer-function-type': 'error',
+          '@typescript-eslint/prefer-readonly': 'error',
```
mightyiam: Personally, I have very little experience writing JavaScript classes, so I'd like some feedback on this, please.

src/index.test.ts
```diff
@@ -60,16 +60,18 @@ test('export', (t): void => {
...
           '@typescript-eslint/promise-function-async': 'error',
           '@typescript-eslint/restrict-plus-operands': 'error',
           '@typescript-eslint/require-array-sort-compare': 'error',
+          '@typescript-eslint/strict-boolean-expressions': 'error',
```
mightyiam: This should be controversial. If you ask many JavaScript developers, they'd tell you that they like boolean casting. I am willing to give it up for the benefit of one less bug once in a while.

src/index.test.ts
```diff
@@ -60,16 +60,18 @@ test('export', (t): void => {
...
+          '@typescript-eslint/triple-slash-reference': ['error', { lib: 'never', path: 'never', types: 'never' }],
```
mightyiam: I have been using TypeScript for a while and never used a triple slash directive. It seems like a thing of the past and perhaps this rule can help keep it in the past.

Yet, precisely because I have never used it, I don't know whether there's a valid use case for it in modern TypeScript. Anyone?

Also, the question of breaking existing codes... The way I see it is that this shareable config is not yet popular and so this is our chance to make great choices, before it gets popular.

src/index.ts
```diff
@@ -66,13 +66,15 @@ export = {
...
-        '@typescript-eslint/no-triple-slash-reference': 'error',
         '@typescript-eslint/no-unnecessary-type-assertion': 'error',
         '@typescript-eslint/no-var-requires': 'error',
         '@typescript-eslint/prefer-function-type': 'error',
+        '@typescript-eslint/prefer-readonly': 'error',
         '@typescript-eslint/promise-function-async': 'error',
         '@typescript-eslint/require-array-sort-compare': 'error',
         '@typescript-eslint/restrict-plus-operands': 'error',
+        '@typescript-eslint/strict-boolean-expressions': 'error',
+        '@typescript-eslint/triple-slash-reference': ['error', { lib: 'never', path: 'never', types: 'never' }],
```
mightyiam: Please take a look at [this comment](https://github.com/typescript-eslint/typescript-eslint/commit/af70a5932f8032cde13f4a0c98d6657cdb40b441#r34796039), by the maintainer of `@typescript-eslint`.

src/index.test.ts
Dangoo: @mightyiam Never used them before myself, but for example major parts of TypeScripts own typings are using that reference style. So maybe not so "out of the past"?

https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

mightyiam: @Dangoo, thank you for your input. Perhaps it was in a podcast that I've heard this—not sure. In TypeScript itself those are slowly replaced with imports. I remember learning that.

mightyiam:
If I don't get a review within a few days, I'm going to do what I feel is right. So this one's going in.

toddbluhm:
I think you are good on all of these rules. The only one I ran into the most issues with is the `strict boolean expressions` simply because anywhere I used `if (someVar) {}` to check for existence was wrong and when I would do something like `someVar = someVar || {}`. Both cases are really just shortcuts for js and certainly can and do lead to unexpected bugs at times. I think when optional chaining and null coalescing get merge into ts, that will fix a lot of these issues with `strict boolean expressions`.

mightyiam:
@toddbluhm thank you for the input!

PR #141, opened 2019-08-30
chore: editorconfig

mightyiam:


mightyiam:
Developer tooling. Merging without review. If anyone has an issue, I will promptly take care of it.

Issue #138, opened 2019-08-29
null vs undefined enforcement?


toddbluhm:
I just wanted to bring this up, not as a huge debate, but more something to consider with typescript standard. Since typescript has types, this is possibly something we could better enforce here rather than in just the plain `standard` utility.

When I was refactoring my OSS project https://github.com/toddbluhm/env-cmd to use this shared config, I ran into the age-old issue of do I check for `null` or do I check for `undefined` or do I check for both when refactoring for the `strict boolean expressions`. Is that something this config could/should take a stance on?

I know for the official typescript repo they only allow `undefined` https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined. Perhaps this shared standard config could take a similar stance?

It looks like someone has already built a plugin for it here: https://www.npmjs.com/package/eslint-plugin-no-null

Now my opinion on how the rule would work if enforced would be essentially to disallow any new usage of `null` in favor of `undefined`, but to still allow 3rd-party libraries and certain built-in utilities like `JSON.parse` to still return `null` types since those cannot be changed.

I am just trying to start this discussion to see what thoughts or opinions are out there. Perhaps this shared config is just too young to take a strong stance either way and needs more traction before tackling such a big debate?

mightyiam:
Thank you for your idea, @toddbluhm.

I would like this project to take a stance, yes. Standard does aim to provide choices, even if they're not clear cut and need to just be made. And by extension, this project.

I personally use `null`, but I really don't mind changing my habit.

> Now my opinion on how the rule would work if enforced would be essentially to disallow any new usage of null in favor of undefined, but to still allow 3rd-party libraries and certain built-in utilities like JSON.parse to still return null types since those cannot be changed.

Sounds reasonable to me. What about when those `null` values go into third party libraries? That's probably edge case, though; I don't remember using a library that must take `null` as input.

For implementation, can [`ban-types`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md) be used?

> Perhaps this shared config is just too young to take a strong stance either way and needs more traction before tacking such a big debate?

I feel that it's better to make decisions before it gets popular. There's precedent in Standard JS where the authors refrain from making changes that would affect most projects. For example, I would really like [dangling commas](https://github.com/standard/standard/issues/586).

toddbluhm:
[...]

PR #147, opened 2019-09-07
feat(no-null): added no-null plugin and rule

toddbluhm:

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Other, please explain:

Added the [no-null](https://github.com/nene/eslint-plugin-no-null) plugin and the `no-null` rule and also updated the readme file to include installing the new plugin. This PR is also intended as a discussion around the topic for possible inclusion. 

Having said that, this plugin could actually be included in the official standard eslint plugin instead of here as it does not rely on type information. If it is determined that it should be included in the official standard eslint plugin my guess is there would need to be a larger discussion in the actual standard repo. The inclusion there might not even be worth considering given the number of projects it would probably break as well. Perhaps it really is only feasible for this typescript version because it is still small, but I would hate to deviate too much from `standard`.

This PR is in reference to #138 

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "eslint": ">=6.0.1",
     "eslint-plugin-import": ">=2.18.0",
     "eslint-plugin-node": ">=9.1.0",
+    "eslint-plugin-no-null": ">=1.0.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
     "@typescript-eslint/eslint-plugin": ">=2.0.0"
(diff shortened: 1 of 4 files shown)


mightyiam:
Thank you, @toddbluhm. I would go in this direction. Yet, there are issues to resolve.

That plugin doesn't seem sufficient. It doesn't have tests. One open issue in it is calling for allowing comparison with `null`. Like `if (foo === null) return`, which seems correct to me.

What we're looking for, perhaps, is a disallowing of not simply having `null` in our code, but of creating `null` values and passing them on to be used in our own code strictly. That might be possible only with TypeScript.

Thinking loud here. There must be a way to use type information to determine whether that `null` value that we created in our code is used strictly in our own code or is also passed on as an argument or as part of an object that is passed as an argument to a function that is not our own code.

What I described feels like uncharted territory to me. Are there TypeScript ESLint rules that are as complicated as what I described, at all?

It sounds possible with a lot of work.

Unless someone surprises me claiming that this is much easier than I thought, I would close this issue as "unlikely to be implemented".

toddbluhm:
Great comments. I can definitely see where you are coming from. I will close this out then for now and I will see what I can dig up or try my hand at creating such a plugin myself.

PR #172, opened 2019-11-15
fix: replace Object.fromEntries with a custom function

toddbluhm:
**What is the purpose of this pull request? (put an "X" next to the item)**

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Added a custom function that accomplishes the same results as `Object.fromEntries`.

`Object.fromEntries` is only supported on the latest version of Nodejs (v12). This change allows this lib to be used on older versions of Nodejs. Typescript does not transpile `Object.fromEntries` because the ts devs consider it to be a runtime feature (https://github.com/microsoft/TypeScript/issues/32803).

maybe this should just be reverted back to strings for simplicity?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,6 +10,13 @@ const equivalents = [
   'quotes'
 ] as const
 
+function fromEntries<T> (iterable: Array<[string, T]>): { [key: string]: T } {
+  return [...iterable].reduce<{ [key: string]: T }>((obj, [key, val]) => {
+    obj[key] = val
+    return obj
+  }, {})
+}
+
 export = {
   extends: 'eslint-config-standard',
   plugins: ['@typescript-eslint'],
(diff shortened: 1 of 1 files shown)


toddbluhm:
We could use `fromentries` but I was trying to avoid adding a new dependency to the project. Still, want me to use that?

mightyiam:
I'm not sure. fromentries doesn't have TypeScript types. I tried figuring them out and gave up. I even looked at [TypeScript's `Object.fromEntries` type](https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2019.object.d.ts) but failed to use those with a function implementation.

At the very least, I feel that the abstracted function should not include the call to the mapping function. It should be provided "entries". That would keep it simpler, I feel. Would you mind doing that?

mightyiam:
#173

toddbluhm:
@mightyiam I used `fromentries` and created a typescript compatible version. It looks/works much better now.

src/index.ts
```diff
@@ -10,6 +10,13 @@ const equivalents = [
   'quotes'
 ] as const
 
+function fromEntries<T> (iterable: Array<[string, T]>): { [key: string]: T } {
+  return [...iterable].reduce<{ [key: string]: T }>((obj, [key, val]) => {
+    obj[key] = val
+    return obj
+  }, {})
+}
+
```
mightyiam: What I had trouble with is having multiple different possible types for the index signature. Here only `string` is used.

This might be losing some type information in comparison with how the `in` operator is used [here](https://github.com/microsoft/TypeScript/blob/5321dcb09f4d37768662afcfc9ffbe1f6b24d0c9/lib/lib.es2019.object.d.ts#L28). But probably good enough, right?

toddbluhm: Yeah, that is what my thinking was. We just need it to be good enough. We are not planning on using any other index types but `string` and this is not exported nor a standalone piece of the library, just a small built-in util 😄

PR #184, opened 2019-11-28
fix: peer dependencies

### LINKED ISSUE #182: Copy Standard's peerDependencies
This package should have the same peerDeps as eslint-config-standard with the addition of the TypeScript plugin. I'm working on this plus a test for it.

### LINKED ISSUE #183: Our `peerDependencies['@typescript-eslint/eslint-plugin']` is out of date
Working on a fix plug test.

mightyiam:
- Fixes #182
- Fixes #183

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,12 +59,12 @@
     "eslint-config-standard": "^14.1.0"
   },
   "peerDependencies": {
-    "eslint": ">=6.0.1",
+    "eslint": ">=6.2.2",
     "eslint-plugin-import": ">=2.18.0",
     "eslint-plugin-node": ">=9.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
-    "@typescript-eslint/eslint-plugin": ">=2.0.0"
+    "@typescript-eslint/eslint-plugin": ">=2.3.3"
   },
   "devDependencies": {
     "@commitlint/cli": "^8.2.0",
@@ -81,6 +81,7 @@
     "eslint-plugin-standard": "^4.0.1",
     "husky": "^3.1.0",
     "npm-run-all": "^4.1.3",
+    "read-pkg-up": "^7.0.0",
     "standard-version": "^7.0.1",
     "tsconfigs": "^4.0.1",
     "typescript": "3.6.4"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,12 +59,12 @@
     "eslint-config-standard": "^14.1.0"
   },
   "peerDependencies": {
-    "eslint": ">=6.0.1",
+    "eslint": ">=6.2.2",
```
mightyiam: Fixes #182. This is the only out-of-sync peer-dependency.

package.json
```diff
@@ -59,12 +59,12 @@
...
     "eslint-plugin-import": ">=2.18.0",
     "eslint-plugin-node": ">=9.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
-    "@typescript-eslint/eslint-plugin": ">=2.0.0"
+    "@typescript-eslint/eslint-plugin": ">=2.3.3"
```
mightyiam: I'm quite certain this should be the same version string as what we use as `devDependency`, because that's how we know which rules we can add.

src/index.test.ts
```diff
@@ -92,3 +94,25 @@ test('export', (t): void => {
...
+      const name = _name as keyof typeof standardPkg.peerDependencies
+      const ourRange = ourPkg.peerDependencies[name]
+      t.is(ourRange, standardRange)
+    })
+})
+
+test('Dependency @typescript-eslint/eslint-plugin', (t) => {
+  const peerDepPluginRange = ourPkg.peerDependencies['@typescript-eslint/eslint-plugin']
+  const devDepPluginRange = ourPkg.devDependencies['@typescript-eslint/eslint-plugin']
+  t.true(peerDepPluginRange.startsWith('>='))
```
mightyiam: Could anyone please remind me why we're using this as range specifier in peerDependencies and not caret?

LinusU: `>=` also allows new major versions, not sure if we want that though 😄

mightyiam:
Thank you, @LinusU!

src/index.test.ts
mightyiam: Thank you, @LinusU. No, I don't feel that we do. We can't support that.

PR #206, opened 2020-01-15
feat: new rule return-await with in-try-catch

### LINKED ISSUE #199: Considering return-await rule
I'd like some input regarding the possible inclusion of the [return-await](https://github.com/typescript-eslint/typescript-eslint/blob/v2.13.0/packages/eslint-plugin/docs/rules/return-await.md) rule, please. I've also [asked upstream](https://github.com/typescript-eslint/typescript-eslint/issues/1378).

mightyiam:
BREAKING CHANGE: new rule return-await with in-try-catch option.

Closes #199.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -125,6 +125,7 @@ test('export', (t): void => {
           '@typescript-eslint/require-array-sort-compare': 'error',
           '@typescript-eslint/require-await': 'error',
           '@typescript-eslint/restrict-template-expressions': ['error', { allowNumber: true }],
+          '@typescript-eslint/return-await': ['error', 'in-try-catch'],
           '@typescript-eslint/space-before-function-paren': ['error', 'always'],
           '@typescript-eslint/strict-boolean-expressions': 'error',
           '@typescript-eslint/triple-slash-reference': ['error', { lib: 'never', path: 'never', types: 'never' }],
--- a/src/index.ts
+++ b/src/index.ts
@@ -102,6 +102,7 @@ export = {
         '@typescript-eslint/require-await': 'error',
         '@typescript-eslint/restrict-plus-operands': ['error', { checkCompoundAssignments: true }],
         '@typescript-eslint/restrict-template-expressions': ['error', { allowNumber: true }],
+        '@typescript-eslint/return-await': ['error', 'in-try-catch'],
         '@typescript-eslint/strict-boolean-expressions': 'error',
         '@typescript-eslint/triple-slash-reference': ['error', { lib: 'never', path: 'never', types: 'never' }],
         '@typescript-eslint/type-annotation-spacing': 'error'


LinusU:
I really think that we should encourage `return await ...` for anything that returns a Promise.

There is an open issue about this explaining why on eslint, I'll try and find it!

mightyiam:
@LinusU thank you for your input. I read all that and while I am half-convinced in the context of JavaScript, I am more convinced by [Brad's words](https://github.com/typescript-eslint/typescript-eslint/issues/1378#issuecomment-568851009) in the context of TypeScript. Did you read that?

LinusU:
Yeah, I read that, and while I agree that refactoring is easier in TypeScript, that was never an argument for me on the JavaScript for me.

What really tips it for me is the proper stack traces. When trying to debug an issue it's very very nice to see the entire stack trace, and not have some frames essentially "missing".

I think that this is quite similar to tail call optimisations which, I believe, still isn't standardised because it would also drop the function from the call stack when doing what's essentially the same thing as here. That was not considered acceptable without a specific syntax opting in to it.

mightyiam:
I understand the pragmatism in preferring one code style over another for the sake of a specific tool working better that way. I also understand the more puristic approach of rejecting that, on principal. Also, that principal has some merit, in case the specific tool happens to evolve to work the same regardless of the code style and then we have legacy style that must be either explained that way or rules to change that are breaking changes. I seem to prefer the latter, more purist approach.

LinusU:
> [...] for the sake of a specific tool working better that way

I don't think that that is the case here though. It's not one tool, but rather the specification of the language. It would be wrong for e.g. the V8 engine to insert that stack if you don't provide `await`.

In fact, the reason that it even works by _not_ adding `await` is because Promises automatically flat maps, e.g. `Promise.resolve(Promise.resolve(1))` is the same as `Promise.resolve(1)`.

When you _omit_ `await` you are really returning a double wrapped `Promise`, it's just that the promise will unwrap that for you automatically.

or in another way:

> [...]  doing `return foo` instead of `return await foo` would actually be what's wrapping it in another `Promise`. If `return ...` in an `async function` desugars to `return Promise.resolve(...)` then writing `return foo` where `foo` is already a promise would be wrapping it in double promises.

mightyiam:
@LinusU thank you for the explanation. Yet, is it only syntactic sugar, or is it different (although perhaps insignificant) logic?

`await` signifies that the promise should be awaited in the current async function, at that position. Who is a linter to say where in the stack it is semantically correct to await a promise? So, the only argument in TypeScript seems to be the stack traces. As you call them, "proper". But why does one call them proper and [another calls them a "red herring"](https://github.com/typescript-eslint/typescript-eslint/issues/1378#issuecomment-568851009)?

I don't have enough experience debugging promise-based code. I have more experience with observables.

I can't decide on this. I'm leaving it open for discussion.

LinusU:
I left a comment here as well: https://github.com/typescript-eslint/typescript-eslint/issues/1378#issuecomment-591642207

But I really really think that we should enable this rule with the `always` option. After using this in my team I've seen that we really appreciate the better stack traces!

mightyiam:
Let's have at least the benefit of in-try-catch and continue from there?

LinusU:
Hmm, `in-try-catch` will actively disallow having it in the `return await` place though. I think a better "in the mean time" would be if it enforced it in try/catch, but didn't care either way outside. Maybe this isn't possible though 🤔

mightyiam:
You're right.

I'd very much like a decision made. It was hard for me to follow the
discussion.

I lean toward 'always' but would like to have the arguments clearly
summarized against each other.

PR #208, opened 2020-01-16
feat: upgrade plugin & parser to v2.10.0

mightyiam:
In rule no-empty-function allow protected and private constructors.


mightyiam:
Thank you, @LinusU .

PR #224, opened 2020-02-04
upgrade parser & plugin to 2.19.0

mightyiam:


src/index.test.ts
```diff
@@ -46,8 +48,10 @@ test('export', (t): void => {
           'space-before-function-paren': 'off',
           '@typescript-eslint/adjacent-overload-signatures': 'error',
           '@typescript-eslint/array-type': ['error', { default: 'array-simple' }],
+          '@typescript-eslint/ban-ts-comment': 'error',
```
LinusU: Hmm, this will be a breaking change for us 😬 

It would also be a bit weird to disable the warning on a per-case basis since it would then always be two "magical" comments:

```typescript
/* eslint-disable-next-line */
// @ts-ignore
foobar()
```

In my opinion, if someone has already made the special comment `@ts-ignore`, they know what they are doing. I'm not sure that the linter should enforce that we never use it.

I find that there are still quite a few legit cases where it has to be used simply because TypeScript can't express everything that JS does at the moment...

-----

sidenote: I'm also not sure about disallowing `!` as we do now. Sometimes it's simply needed because TypeScript haven't understood that we have already checked the condition in some way...

src/index.test.ts
```diff
@@ -86,6 +90,7 @@ test('export', (t): void => {
             }
           ],
           '@typescript-eslint/no-array-constructor': 'error',
+          '@typescript-eslint/no-dupe-class-members': 'error',
```
LinusU: Nice! 👍

src/index.test.ts
```diff
@@ -102,10 +107,12 @@ test('export', (t): void => {
           '@typescript-eslint/no-misused-new': 'error',
           '@typescript-eslint/no-misused-promises': 'error',
           '@typescript-eslint/no-namespace': 'error',
+          '@typescript-eslint/no-non-null-asserted-optional-chain': 'error',
```
LinusU: This won't do anything at the moment since `@typescript-eslint/no-non-null-assertion` makes it so that you can never come here. But it's still nice to enable it, since we might change `@typescript-eslint/no-non-null-assertion` in the future 👍

src/index.test.ts
```diff
@@ -102,10 +107,12 @@ test('export', (t): void => {
...
           '@typescript-eslint/no-non-null-assertion': 'error',
           '@typescript-eslint/no-this-alias': ['error', { allowDestructuring: true }],
           '@typescript-eslint/no-throw-literal': 'error',
           '@typescript-eslint/no-unnecessary-type-assertion': 'error',
+          '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
```
LinusU: Neat 👍

src/index.test.ts
mightyiam: @LinusU is there a `@ts-ignore` where `any` would not be a better alternative?

Regarding [no-non-null-assertion](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-non-null-assertion.md), could you please make that an issue, if you find it a burden?

mightyiam: @LinusU removed it from this pull request.

mightyiam:
Thank you for the review, @LinusU .

src/index.test.ts
LinusU: I actually think that `@ts-ignore` is a much cleaner way to signal something that isn't possible with TypeScript at the moment than just casting to `any`.

- Since `any` already has a legitimate use case it's hard to search your entire codebase for it, and it doesn't stand out as much when you see it.
- `@ts-ignore` lets you very easy add context on _why_ you have to ignore it. Most of the time I will link to a TypeScript issue directly in the comment:
    `// @ts-ignore https://github.com/microsoft/TypeScript/issues/26113`

Thanks for removing it ❤️ I'm open to discussing this further if you want to 👍

PR #229, opened 2020-02-11
feat(rules): new rule @typescript-eslint/ban-ts-comment

mightyiam:
BREAKING CHANGE: new rule @typescript-eslint/ban-ts-comment
https://github.com/typescript-eslint/typescript-eslint/blob/v2.19.2/packages/eslint-plugin/docs/rules/ban-ts-comment.md

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -48,6 +48,7 @@ test('export', (t): void => {
           'space-before-function-paren': 'off',
           '@typescript-eslint/adjacent-overload-signatures': 'error',
           '@typescript-eslint/array-type': ['error', { default: 'array-simple' }],
+          '@typescript-eslint/ban-ts-comment': 'error',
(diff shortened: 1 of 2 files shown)


mightyiam:
@LinusU this new rule is removed from #224, where you [had a reservation](https://github.com/standard/eslint-config-standard-with-typescript/pull/224#pullrequestreview-353752680).

LinusU:
(originally posted in the other thread)

I actually think that `@ts-ignore` is a much cleaner way to signal something that isn't possible with TypeScript at the moment than just casting to `any`.

- Since `any` already has a legitimate use case it's hard to search your entire codebase for it, and it doesn't stand out as much when you see it.
- `@ts-ignore` lets you very easy add context on _why_ you have to ignore it. Most of the time I will link to a TypeScript issue directly in the comment:
    `// @ts-ignore https://github.com/microsoft/TypeScript/issues/26113`

mightyiam:
Obsolete by #285 .

Issue #232, opened 2020-02-18
func-call-spacing and comma-dangle can cause un-fixable error


felskov:
In the following example, I'm getting an error from func-call-spacing:

```ts
const map = new Map<string, () => void>()
```

Auto-fixing results in the following code, which throws an error from comma-dangle:

```ts
const map = new Map<string,() => void>()
```

**What version of this package are you using?**
13.0.0

**What problem do you want to solve?**
Avoid having un-fixable errors using out-of-the-box configuration

**What do you think is the correct solution to this problem?**
Use @typescript-eslint/func-call-spacing instead of eslint built-in func-call-spacing.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/func-call-spacing.md

**Are you willing to submit a pull request to implement this change?**
Yes

LinusU:
> Use @typescript-eslint/func-call-spacing instead of eslint built-in func-call-spacing.

This sounds great, PR welcome 🙌

felskov:
@LinusU  https://github.com/standard/eslint-config-standard-with-typescript/pull/234

PR #234, opened 2020-02-18
feat: use @typescript-eslint func-call-spacing equivalent

felskov:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Replaced eslint default func-call-spacing with @typescript-eslint/func-call-spacing

#232 

No

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -35,6 +35,7 @@ test('export', (t): void => {
           camelcase: 'off',
           'comma-spacing': 'off',
           'default-param-last': 'off',
+          'func-call-spacing': 'off',
           indent: 'off',
           'no-array-constructor': 'off',
           'no-dupe-class-members': 'off',
@@ -66,6 +67,7 @@ test('export', (t): void => {
             allowTypedFunctionExpressions: true,
             allowDirectConstAssertionInArrowFunctions: true
           }],
+          '@typescript-eslint/func-call-spacing': ['error', 'never'],
           '@typescript-eslint/indent': ['error', 2, {
             SwitchCase: 1,
             VariableDeclarator: 1,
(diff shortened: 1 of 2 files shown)


mightyiam:
@felskov thank you for the PR. One question. Did you manually test that this solves #232, please?

felskov:
@mightyiam Yes, I added the following code to `fixture.ts` and ran `npm run lint` prior to and after implementing the change, and can confirm that issues are no longer reported.

```ts
export const map = new Map<string, () => void>()
```

mightyiam:
Another question, please, @felskov. Do you think that this could trigger any new lint errors for users that upgrade? If so, then the commit message should include the "BREAKING CHANGE: blah blah" pattern, please. And it would also be nice to have the "fixes #issuenumber" part in the commit message, please. See commit log for reference.

LinusU:
Great work! 👏 

> Do you think that this could trigger any new lint errors for users who upgrade?

My guess is that it shouldn't, since the previous rule simply didn't understand the extra type parameters 🤔

felskov:
@mightyiam To the best of my knowledge the @typescript-eslint equivalent is a drop-in replacement for the eslint built-in rule, just adding support for syntax the original rule doesn't understand.

Should I amend the commit message to reference the issue and push an update?

mightyiam:
In the docs for the rule it says

> This version adds support for generic type parameters on function calls

Perhaps introducing this rule could trigger linting errors on generic type parameters on function calls?

Seems to me like a breaking change. Like:

```
BREAKING CHANGE: new rule @typescript-eslint/func-call-spacing
```

And yes, @felskov. Please add the "fixes" reference, as well.

felskov:
@mightyiam No problem, I've amended the message as requested

mightyiam:
Thank you so much!

PR #251, opened 2020-03-17
feat: upgrade parser & plugin to v2.24.0

### LINKED ISSUE #247: Upgrade parser & plugin to 2.23.0 (2020-03-09)
https://github.com/typescript-eslint/typescript-eslint/releases/tag/v2.23.0

### LINKED ISSUE #250: Upgrade parser & plugin to 2.24.0 (2020-03-16)
Depends on #247 

https://github.com/typescript-eslint/typescript-eslint/releases/tag/v2.24.0

mightyiam:
BREAKING CHANGE: parser & plugin upgraded to v2.24.0 and new rules.

New rules:

- @typescript-eslint/no-unsafe-call
- @typescript-eslint/no-unsafe-member-access
- @typescript-eslint/no-unsafe-return

Closes #247.
Closes #250.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "TypeScript"
   ],
   "dependencies": {
-    "@typescript-eslint/parser": "^2.22.0",
+    "@typescript-eslint/parser": "^2.24.0",
     "eslint-config-standard": "^14.1.0"
   },
   "peerDependencies": {
@@ -64,14 +64,14 @@
     "eslint-plugin-node": ">=9.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
-    "@typescript-eslint/eslint-plugin": ">=2.22.0"
+    "@typescript-eslint/eslint-plugin": ">=2.24.0"
   },
   "devDependencies": {
     "@commitlint/cli": "^8.2.0",
     "@commitlint/config-conventional": "^8.2.0",
     "@commitlint/travis-cli": "^8.2.0",
     "@types/node": "^13.7.7",
-    "@typescript-eslint/eslint-plugin": "^2.22.0",
+    "@typescript-eslint/eslint-plugin": "^2.24.0",
     "ava": "^3.5.0",
     "editorconfig-checker": "^3.0.4",
     "eslint": "^6.7.2",


src/index.test.ts
```diff
@@ -109,6 +109,9 @@ test('export', (t): void => {
           '@typescript-eslint/no-throw-literal': 'error',
           '@typescript-eslint/no-unnecessary-type-assertion': 'error',
           '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
+          '@typescript-eslint/no-unsafe-call': 'error',
+          '@typescript-eslint/no-unsafe-member-access': 'error',
```
LinusU: Hmmm, this is probably a rather big breaking change, it will basically disallow "tricking" the compiler with `as any`.

On the other hand, I think that `@ts-ignore` or `as any as Foo` are better replacements anyways so I think that I'm pro this change 👍 

We _might_ want to let others chime in first though 🤔

mightyiam: Thank you, @LinusU. It's hard for me to pull the trigger on this, because it won't be my code that is breaking in several places.

LinusU:
@mightyiam should have been more clear with this when I commented, but I think that `no-unsafe-return` is great and shouldn't cause any problem, it's just `no-unsafe-call` and `no-unsafe-member-access` that's a bit complicated.

I looked thru a few of my projects I'm currently working on and `no-unsafe-call` & `no-unsafe-member-access` basically makes it impossible for projects that are currently being converted from JavaScript to TypeScript, as entire parts of the application will be typed as `any`.

All in all, I think that we should add `no-unsafe-return`, but not the other ones ☺️

mightyiam:
What's different about no-unsafe-return? Only a lower rate of occurrence in code?

mightyiam:
Merged without new rules. Making separate pull requests.

LinusU:
> What's different about no-unsafe-return? Only a lower rate of occurrence in code?

The difference is that you should always be able to add an extras `as Foo` since you'll know the return type of the function, e.g.

```typescript
function foo(): Foo {
  return JSON.parse('...') as any
}
```

can easily be written as:


```typescript
function foo(): Foo {
  return JSON.parse('...') as any as Foo
}
```

PR #253, opened 2020-03-17
feat: new rule @typescript-eslint/return-await always

### LINKED ISSUE #199: Considering return-await rule

mightyiam:
BREAKING CHANGE: new rule @typescript-eslint/return-await always

Closes #199.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -125,6 +125,7 @@ test('export', (t): void => {
           '@typescript-eslint/require-array-sort-compare': 'error',
           '@typescript-eslint/require-await': 'error',
           '@typescript-eslint/restrict-template-expressions': ['error', { allowNumber: true }],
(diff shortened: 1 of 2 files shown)


toddbluhm:
I know I am a bit late to the party, but I agree with @LinusU on this. I do think `return await` improves readability a lot and the additional frame in the stack trace is very useful. The performance costs seem to be negligible (and may even be non-existent in production builds/transpiling for typescript).

mightyiam:
@toddbluhm late feedback also helps :smile: thank you.

KilianKilmister:
Some even later feedback: #417

PR #275, opened 2020-04-21
fix: ensure @typescript-eslint/semi rule is applied

bengourley:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Added `@typescript-eslint/semi` to the list of equivalent rules.

Consistent semi-colon usage was not being applied to TypeScript code when linting with this config.

There may be other equivalent packages I haven't noticed. I only added the one I was knowingly being affected by, but there may be a better way to ensure the equivalents don't get out of date.


src/index.ts
```diff
@@ -12,7 +12,8 @@ const equivalents = [
   'no-unused-expressions',
   'no-useless-constructor',
   'quotes',
-  'space-before-function-paren'
+  'space-before-function-paren',
+  'semi'
```
mightyiam, line 16:
```suggestion
  'semi',
  'space-before-function-paren'
```

Just trying to keep this sorted.

PR #285, opened 2020-04-23
feat: rule @typescript-eslint/prefer-ts-expect-error

mightyiam:
BREAKING CHANGE: add rule @typescript-eslint/prefer-ts-expect-error

https://github.com/typescript-eslint/typescript-eslint/blob/v2.29.0/packages/eslint-plugin/docs/rules/prefer-ts-expect-error.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "eslint-plugin-node": ">=9.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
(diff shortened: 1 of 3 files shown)


LinusU:
Just realised that TS 3.9 isn't released yet 😄 

It's going to be released on May 12th though, which is quite soon. We could probably hold of on making another release until then? 🤔

LinusU:
after #287 is merged we should probably rebase this and update the version to `>= 3.9` in this PR

PR #287, opened 2020-04-23
fix: specify typescript as peerDep

### LINKED ISSUE #286: typescript should be a peerDependency
As mentioned here: https://github.com/standard/eslint-config-standard-with-typescript/pull/285#pullrequestreview-398939997.

The parser and plugin do not specify it. Even if they did, ours could possibly be higher?

https://github.com/typescript-eslint/typescript-eslint/blob/4ef67884962b6aac61cc895aaa3ba16aa892ecf4/packages/eslint-plugin/package.json#L58-L61

https://github.com/typescript-eslint/typescript-eslint/blob/4ef67884962b6aac61cc895aaa3ba16aa892ecf4/packages/parser/package.json#L41-L43

mightyiam:
Fixes #286.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "eslint-plugin-node": ">=9.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
+    "typescript": ">=3.8",
     "@typescript-eslint/eslint-plugin": ">=2.29.0"
   },
   "devDependencies": {


package.json
```diff
@@ -68,6 +68,7 @@
     "eslint-plugin-node": ">=9.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "eslint-plugin-standard": ">=4.0.0",
+    "typescript": ">=3.8",
```
mightyiam, line 71: I don't know the version that would actually be necessary. I just picked the latest.

LinusU: Yeah, I think that this is reasonable 👍

PR #311, opened 2020-05-19
chore: commitlint-config-awesome

mightyiam:


mightyiam:
This is to fix that:
https://travis-ci.org/github/standard/eslint-config-standard-with-typescript/jobs/688496463#L219-L221

Issue #303, opened 2020-05-16
[Bug] Configuration for rule "dot-notation" is invalid


mubaidr:
**What version of this package are you using?**
```json
    "eslint": "^7.0.0",
    "eslint-config-prettier": "^6.11.0",
    "eslint-config-standard": "^14.1.1",
    "eslint-config-standard-with-typescript": "^17.0.0",
    "eslint-plugin-import": "^2.20.2",
    "eslint-plugin-jest": "^23.13.0",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-prettier": "^3.1.3",
    "eslint-plugin-promise": "^4.2.1",
    "eslint-plugin-standard": "^4.0.1",
```

**What operating system, Node.js, and npm version?**
Windows: 10, Node.js: 12.16.0, mpn: 6.14.4

**What happened?**
version `16` for `eslint-config-standard-with-typescript` works fine but version `17` throws error: 

```bash
Oops! Something went wrong! :(

ESLint: 7.0.0

Error: .eslintrc.json » eslint-config-standard-with-typescript » eslint-config-standard:
        Configuration for rule "dot-notation" is invalid:
        Value {"allowKeywords":true,"allowPattern":"","allowPrivateClassPropertyAccess":false} should NOT have additional properties.

    at validateRuleOptions (D:\current\brain.js\node_modules\eslint\lib\shared\config-validator.js:132:19)
    at D:\current\brain.js\node_modules\eslint\lib\shared\config-validator.js:187:9
    at Array.forEach (<anonymous>)
    at validateRules (D:\current\brain.js\node_modules\eslint\lib\shared\config-validator.js:184:30)
    at validateConfigArray (D:\current\brain.js\node_modules\eslint\lib\shared\config-validator.js:312:9)
    at CascadingConfigArrayFactory._finalizeConfigArray (D:\current\brain.js\node_modules\eslint\lib\cli-engine\cascading-config-array-factory.js:464:13)
    at CascadingConfigArrayFactory.getConfigArrayForFile (D:\current\brain.js\node_modules\eslint\lib\cli-engine\cascading-config-array-factory.js:275:21)
    at FileEnumerator._iterateFilesRecursive (D:\current\brain.js\node_modules\eslint\lib\cli-engine\file-enumerator.js:467:49)
    at _iterateFilesRecursive.next (<anonymous>)
[...]

PR #314, opened 2020-05-19
fix: [dot-notation] explicitly redefine the rule

andrefilimono:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

The rule **@typescript-eslint/dot-notation** is explicitly defined now.

#303 

No


LinusU:
Hmmm, I don't see how this will fix the problem, have you verified that it does?

As you can see in `src/index.test.ts`, the output hasn't really changed, and we were already using `@typescript-eslint/dot-notation`, and the normal `dot-notation` was turned off...

LinusU:
I added my comment on what I think the problem is here: https://github.com/typescript-eslint/typescript-eslint/issues/2047#issuecomment-631320530

andrefilimono:
I spent some time to debug the internals of `@typescript-eslint`, it works as expected and ESLint applies the correct `@typescript-eslint/dot-notation` rule in runtime.

But when it comes down to the `eslint-config-standard-with-typescript` I see that ESLint uses the wrong rule `dot-notation` instead of `@typescript-eslint/dot-notation`. 
But when I apply changes from this PR it starts using the correct one. 
Actually I don't know why, but it just works.

LinusU:
Hmmm, this is super weird. If I modify both `.eslintrc.json` and `__tests__/.eslintrc.json` to be exactly the same, and then run `npm run lint:eslint` I get the error. If I then remove `__tests__/.eslintrc.json` the error goes away 🤔 

in this repo: https://github.com/BrainJS/brain.js

PR #316, opened 2020-05-20
fix: deep copy standardRules config

### LINKED ISSUE #303: [Bug] Configuration for rule "dot-notation" is invalid

Patrick-Remy:
This fixes #303 invalid configuration for dot-notation

[ ] Documentation update
[x ] Bug fix
[ ] New feature
[ ] Other, please explain:

Instead of assigning the equivalents config object, the must be deep copied

Addresses #303

more elegant way for `[...val] as unknown as T`


LinusU:
Awesome! Thank you for finding the underlying cause of the issue, this is hilarious 😁 

What do you think about this approach for fixing it instead though:

```diff
diff --git a/src/index.ts b/src/index.ts
index a6254b5..05faf51 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,6 +26,10 @@ function fromEntries<T> (iterable: Array<[string, T]>): { [key: string]: T } {
   }, {})
 }
 
+function clone<T extends object | string | number | boolean | null> (value: T): T {
+  return JSON.parse(JSON.stringify(value))
+}
+
 export = {
   extends: 'eslint-config-standard',
   plugins: ['@typescript-eslint'],
@@ -44,7 +48,7 @@ export = {
         'no-use-before-define': 'off',
 
         // @typescript-eslint versions of Standard.js rules:
-        ...fromEntries(equivalents.map((name) => [`@typescript-eslint/${name}`, standardRules[name]])),
+        ...fromEntries(equivalents.map((name) => [`@typescript-eslint/${name}`, clone(standardRules[name])])),
         '@typescript-eslint/no-use-before-define': ['error', {
           functions: false,
           classes: false,
```

The benefits are:

1. It actually does a deep clone, your code currently does a shallow clone
2. It keeps the behaviour of `fromEntries` aligned with the official `Object.fromEntries` (which we can switch to when targeting newer Node.js version)
3. Avoids cloning the newly created `[name, 'off']` array

Patrick-Remy:
Also just thought about that. I updated the commit 👍

Patrick-Remy:
Yes, sorry for the history files. I added an extensive test for it 👍

src/index.test.ts
```diff
@@ -192,3 +193,32 @@ test('Deps parser and plugin are same version', async (t) => {
...
+      t.not(ruleConfig, comparisonRules[comparisonRuleName])
+    }
+  }
+}
+
+test('No references in rules config', (t) => {
+  for (const override of exported.overrides) {
+    const rules: RulesConfig = override.rules
+    // Rules must not have object references inside theirselves
+    testNotHaveObjectReferences(t, rules, rules, true)
```
mightyiam, line 220: Drop this part of the test.

src/index.test.ts
```diff
@@ -192,3 +193,32 @@ test('Deps parser and plugin are same version', async (t) => {
...
+    // Rules must not have object references to standard rules
+    testNotHaveObjectReferences(t, rules, standardRules)
```
mightyiam, line 222: Inline this part, instead of a function call. And drop the comment.

src/index.test.ts
```diff
@@ -192,3 +193,32 @@ test('Deps parser and plugin are same version', async (t) => {
...
+    for (const comparisonRuleName in comparisonRules) {
+      if (skipSameKey && ruleName === comparisonRuleName) continue
+      t.not(ruleConfig, comparisonRules[comparisonRuleName])
+    }
+  }
+}
+
+test('No references in rules config', (t) => {
+  for (const override of exported.overrides) {
+    const rules: RulesConfig = override.rules
```
mightyiam, line 218: No need for this type, then, I guess.

src/index.test.ts
```diff
@@ -192,3 +193,32 @@ test('Deps parser and plugin are same version', async (t) => {
...
+ */
+function testNotHaveObjectReferences (t: ExecutionContext<unknown>, rules: RulesConfig, comparisonRules: RulesConfig, skipSameKey: boolean = false): void {
+  for (const ruleName in rules) {
+    const ruleConfig = rules[ruleName]
+    if (typeof ruleConfig !== 'object') continue // Non-objects use copy-by-value
+
+    for (const comparisonRuleName in comparisonRules) {
+      if (skipSameKey && ruleName === comparisonRuleName) continue
+      t.not(ruleConfig, comparisonRules[comparisonRuleName])
+    }
```
mightyiam, line 212: Instead of this iteration, check only against the same key.

Patrick-Remy:
I think the tests are both required. 

The iteration is required, as the error can occur, if there are any references, also from different keys. You can validate it by following these steps:

(for testing I used the compiled .js files):
1. Create a new project with this `package.json`
```json
{
  "dependencies": {
    "@types/node": "^14.0.5",
    "@typescript-eslint/eslint-plugin": "^3.0.0",
    "@typescript-eslint/parser": "^3.0.0",
    "eslint": "^7.1.0",
    "eslint-config-standard": "^14.1.1",
    "eslint-plugin-import": "^2.20.2",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-promise": "^4.2.1",
    "eslint-plugin-standard": "^4.0.1",
    "typescript": "^3.9.3"
  }
}
```
this `tsconfig.json`
```json
{
  "compilerOptions": {
    "target": "es2018",
    "baseUrl": "src",
    "outDir": "lib",
    "module": "commonjs"
  },
  "include": [
    "src"
  ]
}
```
and this `.eslintrc.js`
```js
const eslintrc_json_1 = require("eslint-config-standard/eslintrc.json");
module.exports = {
  extends: 'eslint-config-standard',
  plugins: ['@typescript-eslint'],
  overrides: [
    {
      files: ['*.ts', '*.tsx'],
      parser: '@typescript-eslint/parser',
      rules: {
        '@typescript-eslint/dot-notation': eslintrc_json_1.rules['arrow-spacing']
      }
    }
  ]
}
```
2. Run `npm install`
3. Now edit `node_modules/eslint-config-standard/eslintrc.json` and replace the rule config for `arrow-spacing` with `["error", { "allowKeywords": true }]`.
4. Open `eslint-test/node_modules/eslint/lib/rules/arrow-spacing.js` and add 
```js
allowKeywords: {
    type: 'boolean',
    default: true,
},
allowPattern: {
    type: 'string',
    default: '',
}
```
to the `schema` properties.
5. Now run `npx eslint src` and see this error 
```
Error: .eslintrc.js#overrides[0]:
        Configuration for rule "@typescript-eslint/dot-notation" is invalid:
        Value {"allowKeywords":true,"before":true,"after":true,"allowPattern":"","allowPrivateClassPropertyAccess":false} should NOT have additional properties.
```

So `before` and `after` default properties have been applied to the `arrow-spacing` config object, and therefor also to the `@typescript-eslint/dot-notation` rule.
I think the problem comes from default values and eslint's schema, but haven't got as deep in eslint's code. I also tested it with different rules where config schema is already the same, but without success. Therefor the above modification of the schema is required, because no other package has the same schema as `@typescript-eslint/dot-notation`.

----

References inbetween the override could also be problematic. Just imagine my personal config extends `eslint-config-standard-with-typescript`, and imports it to modificate it like
```js
const config = require('eslint-config-standard-with-typescript')

config.overrides[0].rules['some-rule'] = ['error', {
  custom: false
}]

module.exports = config
```

If this lib would use the same config object for 'some-rule' and 'some-different-rule', this will result in unexpectedly modifying both of them. So it would be a bad style to use references inside the config.



Hope you can get my points.

mightyiam:
@Patrick-Remy the only reuse-of-rule-by-reference we have is same-key from eslint-config-standard. That is removed in this PR and that is the only thing that should be tested against in this PR. More comprehensive testing is not worth the slightly increased complexity of the test, because there is no reasonable expectation that what it covers will ever be actually hit. Please do make those changes so that we could make a release.

src/index.test.ts
Patrick-Remy: It is required due to `Element implicitly has an 'any' type because expression of type 'string' can't be used to index type` if use `standardRules[ruleName]`.

Patrick-Remy:
In my opinion tests should not use knowledge about the implementation details, to only cover that problems never reoccur in a specific case. The test now isn't preventing the root cause in a general way any more. It also hadn't any big performance downsides and it's now possible to pass the test, but run into this error again in a different case. But possibly that's only about philosophy.

However, I also think a fast release is important now, as the package is broken for the most. If there are any changes required, please let me now.

mightyiam:
Thank you, @Patrick-Remy :blue_heart:

PR #359, opened 2020-06-11
feat: require-array-sort-compare ignoreStringArrays

mightyiam:


mightyiam:
I feel that native string sorting is sometimes useful and there's not enough justification to remind people that it might not be what they're looking for. Not with a rule.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/require-array-sort-compare.md#ignorestringarrays

Divlo (approved):

mightyiam:
Seems to require a rebase?

Divlo:
Yep, we need to squash maybe, so `commitlint` is happy.
I updated the branch with master but `commitlint` doesn't like that.

PR #529, opened 2021-02-25
chore: update husky

Divlo:

- [x] Update husky from `v4` to `v5` major version bump.

closes #516

Here is the link for the migration from `v4` to `v5` : https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v5

`pinst` is made by the same creator of `husky`, it allows to only run the postinstall hook in development, typically, when our user install `eslint-config-standard-with-typescript`, it doesn't make sense that they install/setup husky, it only make sense, when we're developing the package.
Link: https://github.com/typicode/pinst


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
mightyiam, line 1: I may have encountered this issue before, so here's a guess.
If the purpose of this underscore is to please editorconfig, then please convince editorconfig otherwise, instead, by editing `.editorconfig`, adding a section for this file specifically, setting it to not have a closing newline.

mightyiam (commented):
Just need to know that this is manually tested and confirmed to fail a non-conforming commit message.

.husky/.gitignore
Divlo: Husky `v5` add a `_` folder with a `sh` script, and `husky` ignore it.
I didn't create this `.gitignore`, it is automatically managed/generated by husky since `v5`.

mightyiam: Oh, I see.

Divlo:
I can confirm that it is working.

mightyiam (approved):
OK. Please squash and merge.

PR #530, opened 2021-02-25
ci: change Travis to GitHub actions

### LINKED ISSUE #598: CI no longer running
> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

Divlo:

- [x] Change travis configuration to GitHub Actions for continuous integration

Fixes #598


.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [master]
+
+jobs:
+  commitlint:
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v2'
+        with:
+          fetch-depth: 0
+      - uses: 'wagoid/commitlint-github-action@v3'
```
mightyiam, line 18: Could the `configFile` be specified, please? I know the default is correct. I prefer the clarity.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: 'ubuntu-latest'
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
```
mightyiam, line 15: Currently, we have renovate updating these versions. In this setup, will we lose this convenience? I wouldn't like losing it.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: 'actions/setup-node@v2.1.5'
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: 'Cache Node dependencies'
+        uses: 'actions/cache@v2.1.4'
+        with:
+          path: '~/.npm'
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+          restore-keys: ${{ runner.os }}-node-
```
mightyiam, line 34: Does this decrease the time of the next step? I don't like that I have to dig into some documentation just to figure out what the meaning of the `key` is and what the value of it here does. So I didn't. How much difference does this make?

.travis.yml
```diff
@@ -1,11 +0,0 @@
-language: node_js
-node_js:
-  - "10"
-  - "12"
-  - "14"
-  - "15"
-before_install:
-  - npm --global install npm
```
mightyiam, line 8: This step should be preserved, so that we use the latest version of npm in CI.

mightyiam (changes requested):

.github/workflows/commitlint.yml
Divlo: Oh right, I also prefer clarity, actually I didn't know, we could provide `configFile`, I just saw on the doc : https://github.com/marketplace/actions/commit-linter

Thanks, I'll also update my repos with that.

.travis.yml
Divlo: Sure, done!

.github/workflows/nodejs.yml
Divlo: Basically, it allows to cache `node_modules`, so yes it saves some time, I think it is better, so we have a faster CI.

.travis.yml
Divlo: Actually, as said, `npm@7` fails with invalid `peerDependencies`, so that's why the CI is failing currently, but once #531 has been merged, everything will work fine! :smile:

.github/workflows/nodejs.yml
Divlo: See: https://github.com/standard/eslint-config-standard-with-typescript/pull/533#issuecomment-785785803

mightyiam: As I responded there—I wouldn't like to lose this convenience.

LinusU: Hmmm, sorry if I'm missing something, but if renovate is updating these versions won't we stop testing on a specific version of Node.js without cutting a breaking change? As I see it, we should only remove older Node.js versions here when we drop support for them by changing `"engines"` in `package.json`, and release a new major version? 🤔

mightyiam: If renovate removes a Node.js version, that PR should not be automerged. It should be manually merged, having its commit message changed to include the removal of support. I don't remember whether renovate currently auto-merges these particular PRs. And I don't remember the last time this had happened.

Divlo:
See: https://github.com/standard/eslint-config-standard-with-typescript/pull/533#issuecomment-813521397

.github/workflows/nodejs.yml
Divlo: Actually, we don't need it. It makes things simpler, if we really want to use cache, we should use `package-lock.json` but that would be for another PR, no need for this one. :+1:

.github/workflows/nodejs.yml
Divlo: Thanks to @voxpelli note in https://github.com/standard/eslint-config-standard-with-typescript/issues/598#issuecomment-871624204.

We can now use `lts/*` for `node-version`, so we still support all the versions of Node.js.

Divlo:
Reopening to fix issue #598.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+    branches: [master]
+
+jobs:
+  ci:
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v2'
+        with:
+          fetch-depth: 0
```
mightyiam, line 15: Is this for commit linting?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - uses: 'wagoid/commitlint-github-action@v3'
+        with:
+          configFile: 'commitlint.config.js'
+
+      - name: 'Use Node.js'
+        uses: 'actions/setup-node@v2.2.0'
+        with:
+          node-version: 'lts/*'
```
mightyiam, line 24: Does this mean there will be a matrix of all the Node.js versions?

mightyiam: I would prefer GitHub Actions but while Travis currently runs in all supported Node.js versions, this seems to use only the latest LTS. Is there a nice way to make this run all supported Node.js releases without needing to update those versions manually?

mightyiam (changes requested):

.github/workflows/ci.yml
Divlo: Yes.

.github/workflows/ci.yml
Divlo: You are right, it is only using the latest LTS version, I thought otherwise.
I opened a PR to `actions/setup-node`, to support the `current` alias : https://github.com/actions/setup-node/pull/279

mightyiam: @Divlo would a `current` feature in that action allow us to generate a matrix of several Node.js versions? That is what Travis CI has been doing for us.

Divlo: Yes, it will allow us to use `node-version: [lts/*, current]`, so it will run the CI for the latest LTS Node.js version (currently v14) and latest Current Node.js version (currently v16).

voxpelli: To generate a matrix, a strategy like this one would have to be used: https://github.com/es-shims/Promise.prototype.finally/blob/0bbf745cf4b427aee96262a59bf369b268d6995d/.github/workflows/node-native.yml#L6-L25

Else `[10, 12, lts/*, current]` will leave gaps when a new major lts or current release is made.

voxpelli: Reference: https://github.com/actions/setup-node/issues/26#issuecomment-773769435

mightyiam: So assuming a matrix is supported, I'd like an additional step that reads the Node.js versions from a file in the repository that can be updated via Renovate. It would be rather hacky, but I wouldn't mind if this file is the `.travis.yml` even though we wouldn't be using Travis CI. Of course, a comment in that file explaining its existence would be in order.

Divlo: Yes, matrix is supported.

> I'd like an additional step that reads the Node.js versions from a file in the repository that can be updated via Renovate

I'm not against this, I guess you read that from https://github.com/renovatebot/renovate/issues/7716.
But I would rather have `node-version: [lts/*, current]` if my PR get merged in `actions/setup-node`.

> I wouldn't mind if this file is the `.travis.yml` even though we wouldn't be using Travis CI.

I don't think that's a good idea, we should not use a `.travis.yml` file if we are not using Travis CI.

We will see what we can do, but first we should make sure that the CI pass again and currently it is not the case (see: https://github.com/standard/eslint-config-standard-with-typescript/pull/597).

Divlo:
This PR has no recent activity, and probably won't be merged for a long time and as I don't want/have time to do the job for this PR, I decided to close it.

Feel free to open another PR yourself based on this one if you want this to be implemented.

mightyiam:
Travis CI is not as kind as GitHub Actions with regards to how many runs/minutes the project gets. We have incentive to switch to Actions.

PR #531, opened 2021-02-25
chore: usage of @commitlint/config-conventional

Divlo:

- [x] Change `commitlint-config-awesome` to `@commitlint/config-conventional`
`commitlint-config-awesome` is less used/updated (not updated since Dec 16, 2019), break on `npm@7` because peerDependency don't match and we don't need it anymore because we don't use anymore `Greenkeeper`, also `@commitlint/config-conventional` is what it is used for `ts-standard`.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,1 +1,1 @@
-module.exports = { extends: ['awesome'] }
+module.exports = { extends: ['@commitlint/config-conventional'] }
--- a/package.json
+++ b/package.json
@@ -59,12 +59,12 @@
     "eslint-config-standard": "^16.0.0"
   },
   "peerDependencies": {
-    "@typescript-eslint/eslint-plugin": ">=4.0.1",
     "eslint": ">=7.12.1",
     "eslint-plugin-import": ">=2.22.1",
     "eslint-plugin-node": ">=11.1.0",
     "eslint-plugin-promise": ">=4.2.1",
-    "typescript": ">=3.9"
+    "typescript": ">=3.9",
+    "@typescript-eslint/eslint-plugin": ">=4.0.1"
   },
   "devDependencies": {
     "@commitlint/cli": "11.0.0",
--- a/package.json
+++ b/package.json
@@ -59,21 +59,21 @@
     "eslint-config-standard": "^16.0.0"
   },
   "peerDependencies": {
+    "@typescript-eslint/eslint-plugin": ">=4.0.1",
     "eslint": ">=7.12.1",
     "eslint-plugin-import": ">=2.22.1",
     "eslint-plugin-node": ">=11.1.0",
     "eslint-plugin-promise": ">=4.2.1",
-    "typescript": ">=3.9",
-    "@typescript-eslint/eslint-plugin": ">=4.0.1"
+    "typescript": ">=3.9"
   },
   "devDependencies": {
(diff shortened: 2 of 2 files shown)


commitlint.config.js
```diff
@@ -1 +0,0 @@
-module.exports = { extends: ['awesome'] }
```
mightyiam, line 1: Keep the config in this file, unless there's a specific reason to change it.

package.json
```diff
@@ -71,13 +71,12 @@
     "@typescript-eslint/eslint-plugin": ">=4.0.1"
   },
   "devDependencies": {
-    "@commitlint/cli": "11.0.0",
```
mightyiam, line 74: Don't update the version unless you have to, for the purpose of this PR (which is about replacing `commitlint-config-awesome` with `@commitlint/config-conventional`.

package.json
```diff
@@ -71,13 +71,12 @@
...
-    "@commitlint/travis-cli": "11.0.0",
```
mightyiam, line 75: Don't remove this package, because it's still necessary. It's removal should be part of #530.

mightyiam (changes requested):

commitlint.config.js
Divlo: I forgot to tell you why did I change this config file, I could tell you why in "Is there anything you'd like reviewers to focus on?" section of the PR.

Anyway here is the reason : I like consistency, and currently we have `.eslintrc.json`, so it feels better to do the same thing for this config file. If possible, we should try to be consistent in the codebase or for config files.

package.json
Divlo: I don't think it is viable to open another PR just to update a version, it is still related to this PR even if it is indeed not necessary or for the purpose of this PR.

package.json
Divlo: Sure, done! :smile:

commitlint.config.js
mightyiam: Perhaps. But this is not the PR for that change.

package.json
mightyiam: I disagree. Related is not enough. PRs in this project are cheap. If you don't feel like making another PR—don't. But please don't include it with this one.

Divlo: Right, it's up to you! I will open a new PR if needed! :smile:

Divlo:
Hey! :wave: @mightyiam 
Thanks again for taking your time to review my changes and excuse me for this long delay, I just updated to address your comments! Is there is anything that need to change before merging this PR ?

package.json
```diff
@@ -59,21 +59,21 @@
     "eslint-config-standard": "^16.0.0"
   },
   "peerDependencies": {
+    "@typescript-eslint/eslint-plugin": ">=4.0.1",
```
mightyiam, line 62: Was this moved there automatically?

Divlo: Yes when I ran `npm install`.

mightyiam: That's an unrelated change, then. And it should be made prior to any other change made with npm. As a separate PR. No apologies from me on this. This is how I roll.

Divlo: Well, others people, when they'll do PRs will probably change that too I guess.
But fine like that, we'll see, I updated, so it's doesn't change these lines. :+1:

mightyiam:
Thank you, @Divlo.

PR #532, opened 2021-02-25
build(deps): update latest version

Divlo:

- [x] Update dependencies to latest version

Closes #419 

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "eslint-plugin-node": ">=11.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "typescript": ">=3.9",
-    "@typescript-eslint/eslint-plugin": ">=4.0.1"
+    "@typescript-eslint/eslint-plugin": ">=4.15.2"
   },
   "devDependencies": {
     "@commitlint/cli": "11.0.0",
--- a/package.json
+++ b/package.json
@@ -59,8 +59,8 @@
     "TypeScript"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -59,23 +59,23 @@
...
+    "eslint-config-standard": "^16.0.2"
   },
   "peerDependencies": {
     "eslint": ">=7.12.1",
     "eslint-plugin-import": ">=2.22.1",
     "eslint-plugin-node": ">=11.1.0",
     "eslint-plugin-promise": ">=4.2.1",
     "typescript": ">=3.9",
-    "@typescript-eslint/eslint-plugin": ">=4.0.1"
+    "@typescript-eslint/eslint-plugin": ">=4.15.2"
```
mightyiam, line 71: We should not bump this without a good reason. It's a breaking change. A good reason would be using a new rule.

mightyiam (changes requested):

package.json
Divlo: Indeed, do you know how we could update `@typescript-eslint/eslint-plugin` in `devDependencies` without changing the version in `peerDependencies`, because it seems like if we do it breaks the CI.

mightyiam: The test that would be the most useful is one that proves that the peer dep of the plugin starts with `>=` and is the minimal version that includes all of the rules that we're using, and all of the specific rule configuration options and values that we're setting.

Divlo:
Thanks for your time, I'll close this PR and let `renovabot` do his job! :smile:

PR #533, opened 2021-02-25
ci: change renovate to dependabot

Divlo:

- [x] Change `renovatebot` to `dependabot`
As it is better integrated with GitHub, and if we choose to merge #530, then we would be able to update GitHub Actions as well.

We should remove the stale branches from the old Greenkeeper bot and renovatebot once this has been merged.


mightyiam (changes requested):
Currently, renovate and travis are coupled in this way: renovate updates the Node.js versions in the travis config. I'd like to keep this convenience.

I don't this PR preserves that—correct?

Divlo:
@mightyiam I don't know if it is possible, to update the `node.js` version in strategy matrix in GitHub Actions but I don't think it is really necessary, `Node.js` release a LTS version every 2 years or so and we should always support at least the latest `LTS` version, also I think it would be cool to have this in `package.json` :
```json
  "engines": {
    "node": ">=10"
 },
 ```
 
So we know exactly what `node` version we support, but if we do that, I don't think `renovabot` will update it in `package.json` so it's not a big deal that `renovabot` update the `nodejs` version in `travis` or `GitHub Actions`.

**Comment by mightyiam** [comment]
> @mightyiam I don't know if it is possible, to update the `node.js` version in strategy matrix in GitHub Actions but I don't think it is really necessary, `Node.js` release a LTS version every 2 years or so and we should always support at least the latest `LTS` version, also I think it would be cool to have this in `package.json` :
> 
> ```json
>   "engines": {
>     "node": ">=10"
>  },
> ```
> 
> So we know exactly what `node` version we support, but if we do that, I don't think `renovabot` will update it in `package.json` so it's not a big deal that `renovabot` update the `nodejs` version in `travis` or `GitHub Actions`.

Currently, we support all supported Node.js releases. No reason to degrade from this.

And the CI tries to test this support. That's great. We should switch to GitHub Actions only if we can maintain this feature.

**Comment by Divlo** [comment]
>Currently, we support all supported Node.js releases. No reason to degrade from this.
>
> And the CI tries to test this support. That's great. We should switch to GitHub Actions only if we can maintain this feature.

Right, I'll close this PR and the other one about GitHub Actions, if you rather to stick with renovabot and Travis CI! :+1:

**Comment by Divlo** [comment]
> Currently, we support all supported Node.js releases. No reason to degrade from this.
> 
> And the CI tries to test this support. That's great. We should switch to GitHub Actions only if we can maintain this feature.

Good news, with the new `action/setup-node` version (ref: https://github.com/standard/eslint-config-standard-with-typescript/issues/598#issuecomment-871624204), we can support all the Node.js LTS versions automatically, so we don't degrade with this PR. @mightyiam

**Comment by mightyiam** [comment]
> > Currently, we support all supported Node.js releases. No reason to degrade from this.
> > And the CI tries to test this support. That's great. We should switch to GitHub Actions only if we can maintain this feature.
> 
> Good news, with the new `action/setup-node` version (ref: [#598 (comment)](https://github.com/standard/eslint-config-standard-with-typescript/issues/598#issuecomment-871624204)), we can support all the Node.js LTS versions automatically, so we don't degrade with this PR. @mightyiam

I'm not certain that's true. Discuss in #530 .

File: .github/dependabot.yml [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
```
voxpelli, line 13: Is this the interval we use in all projects? `weekly` feels like a much preferred schedule to me

voxpelli (commented):
> As it is better integrated with GitHub

I would actually personally say that Renovate is better integrated with GitHub and better overall, so I see no urgency or even need in this one

.github/dependabot.yml
Divlo: Yes, it is what we use in other repositories of the Standard organization, example: `standard/standard`: https://github.com/standard/standard/blob/master/.github/dependabot.yml.

> I would actually personally say that Renovate is better integrated with GitHub and better overall

How true is that ? What is missing in `dependabot` that `Renovate` has ?

I would say it is better to stay consistent with others repos of the org and use `dependabot` instead of `renovabot`.

voxpelli: > How true is that ? What is missing in `dependabot` that `Renovate` has ?

Hard to list it all here, check one of my repos:

* https://github.com/voxpelli/eslint-config/pulls
* https://github.com/voxpelli/async-htm-to-string/pulls

And my config:

* https://github.com/voxpelli/renovate-config-voxpelli/blob/3fea9a0cd461d3e9d4eda3c186e83d6d221562d4/package.json#L18

> I would say it is better to stay consistent with others repos of the org and use `dependabot` instead of `renovabot`.

I would generally agree, though the urgency of it is low compared to getting eg. CI to work.

+ if there may be a possibly better alternative, then it might be a good idea to try that out on some repos, and if superior, then move the rest to it, else replace it with the common one.

Divlo:
This PR has no recent activity, and probably won't be merged for a long time and as I don't want/have time to do the job for this PR, I decided to close it.

Feel free to open another PR yourself based on this one if you want this to be implemented.

mightyiam:
@Divlo renovate no longer updates Node.js versions in all of the files where they are stored in this repo. So that's no longer a reason to stick with Renovate.

PR #590, opened 2021-06-14
test: fix issue with peerDeps checking with semver

Divlo:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain: Updated the test, so it passes again

Since `eslint-config-standard@16.0.3`, the test fails because the `peerDependencies` is like this `"eslint-plugin-promise": "^4.2.1 || ^5.0.0"`, and we were checking the strict equality, but instead we should use [semver](https://www.npmjs.com/package/semver) to check if it satisfies the `peerDependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "@commitlint/travis-cli": "12.1.4",
     "@types/eslint": "7.2.13",
     "@types/node": "14.17.3",
+    "@types/semver": "7.3.6",
     "@typescript-eslint/eslint-plugin": "4.0.1",
     "ava": "3.15.0",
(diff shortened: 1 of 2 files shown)


mightyiam (changes requested):
I think the point is that they're identical rather than one satisfying the other. #596.

mightyiam:
So, to be clear, #596 probably supercedes the (unfiled) issue that this PR is attempting to solve.

Divlo:
Closing as #596 will close the issue possibly fixed thanks to this PR.

PR #597, opened 2021-06-19
fix(package): peerDeps use caret ranges

### LINKED ISSUE #596: peerDeps should use `^` range type
I don't know why we're using `>=`. That claims support for larger major versions, which isn't a reasonable claim. The change seems to come from 6c8c04c3a8302df57c751156af24d59c3eca78a3, which was me, but it doesn't explain much other than referring to #94 which also doesn't explain much. I feel that the peerDeps should use `^`. That's a breaking change.

mightyiam:
Fixes #596.

BREAKING CHANGE: peer dependencies are specified with `^`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,12 +59,12 @@
     "eslint-config-standard": "^16.0.0"
   },
   "peerDependencies": {
-    "eslint": ">=7.12.1",
-    "eslint-plugin-import": ">=2.22.1",
-    "eslint-plugin-node": ">=11.1.0",
-    "eslint-plugin-promise": ">=4.2.1",
-    "typescript": ">=3.9",
-    "@typescript-eslint/eslint-plugin": ">=4.0.1"
+    "eslint": "^7.12.1",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^4.2.1",
+    "typescript": "^3.9",
+    "@typescript-eslint/eslint-plugin": "^4.0.1"
   },
   "devDependencies": {
     "@commitlint/cli": "12.1.4",
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -175,7 +175,7 @@ test('Dependencies range types', async (t) => {
     t.true(range.startsWith(specifier), `Regular dependency ${name} starts with \`${specifier}\`.`)
   }
   for (const [name, range] of Object.entries(ourPeerDeps)) {
-    const specifier = '>='
+    const specifier = '^'
     t.true(range.startsWith(specifier), `Peer dependency ${name} starts with \`${specifier}\`.`)
   }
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -59,12 +59,12 @@
...
-    "eslint": ">=7.12.1",
-    "eslint-plugin-import": ">=2.22.1",
-    "eslint-plugin-node": ">=11.1.0",
-    "eslint-plugin-promise": ">=4.2.1",
-    "typescript": ">=3.9",
-    "@typescript-eslint/eslint-plugin": ">=4.0.1"
+    "eslint": "^7.12.1",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^4.2.1",
```
Divlo, line 65: We should support the same `peerDependencies` as [eslint-config-standard](https://github.com/standard/eslint-config-standard) does.

Here we're not supporting`"eslint-plugin-promise": "^5.0.0"` but [eslint-config-standard](https://github.com/standard/eslint-config-standard) does `"eslint-plugin-promise": "^4.2.1 || ^5.0.0"`.

Divlo (changes requested):
Overall looks good to me! :+1: 
But, I think we should support `eslint-plugin-promise@5` and superior, so `"eslint-plugin-promise": "^5.0.0"`.

But if we do this, we don't support anymore `eslint-plugin-promise@4`, so I would recommend to use the same `peerDependencies` as [eslint-config-standard](https://github.com/standard/eslint-config-standard/) : `"eslint-plugin-promise": "^4.2.1 || ^5.0.0"`.

package.json
mightyiam: So how come the test is passing?

mightyiam: The current dependency, `eslint-config-standard@16` doesn't yet have that wider `eslint-plugin-promise` peer dep. So that issue will become relevant in the PR that upgrades the `eslint-plugin-promise` dependency.

Divlo: The test is actually failing on my local machine.
image

Also, it seems like, the CI with Travis doesn't seem to work or is really slow to start...
Are you sure, you don't want to migrate to GitHub Actions?

mightyiam: #598

voxpelli: I agree with @Divlo that it should be `"eslint-plugin-promise": "^4.2.1 || ^5.0.0"` here

```suggestion
    "eslint-plugin-promise": "^4.2.1 || ^5.0.0",
```

Else it would be a regression in what versions are supported + `"eslint-config-standard": "^16.0.0"` has been widened to indicate support for it.

package.json
```diff
@@ -59,12 +59,12 @@
...
+    "typescript": "^3.9",
```
voxpelli, line 66: More common to write:

```suggestion
    "typescript": "^3.9.0",
```

PR #602, opened 2021-07-02
maintenance

mightyiam:
- test(deps): log dep in peer dep inclusion test
- fix(deps): expand eslint-plugin-promise peerDep to equal upstream
- fix(deps): sort peer deps
- chore(deps): update some dev deps


src/index.test.ts
```diff
@@ -191,7 +191,7 @@ test('Own peerDependencies include those of eslint-config-standard', async (t) =
       // https://github.com/microsoft/TypeScript/pull/12253
       const name = _name as keyof typeof standardPkg.peerDependencies
       const ourDep = ourPeerDeps[name]
-      t.is(ourDep, standardDep)
+      t.is(ourDep, standardDep, name)
```
mightyiam, line 194: This shows us which peer dep is not in alignment with upstream.

package.json
```diff
@@ -62,7 +62,7 @@
     "eslint": "^7.12.1",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-promise": "^4.2.1",
+    "eslint-plugin-promise": "^4.2.1 || ^5.0.0",
```
mightyiam: This aligns with upstream.

package.json
```diff
@@ -59,12 +59,12 @@
...
   },
   "peerDependencies": {
+    "@typescript-eslint/eslint-plugin": "^4.0.1",
     "eslint": "^7.12.1",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1 || ^5.0.0",
-    "typescript": "^3.9",
-    "@typescript-eslint/eslint-plugin": "^4.0.1"
+    "typescript": "^3.9"
```
mightyiam: I noticed that npm sorted this for me when I used `npm install` so I gave the sorting its own commit.

package.json
```diff
@@ -71,21 +71,21 @@
     "@commitlint/config-conventional": "12.1.4",
     "@commitlint/travis-cli": "12.1.4",
     "@types/eslint": "7.2.13",
-    "@types/node": "14.17.3",
+    "@types/node": "15.14.0",
```
mightyiam, line 74: I'm not sure how versioning for this works but as long as type checking works we're good, I guess.

package.json
```diff
@@ -71,21 +71,21 @@
...
     "@typescript-eslint/eslint-plugin": "4.0.1",
     "ava": "3.15.0",
     "editorconfig-checker": "4.0.2",
-    "eslint": "7.28.0",
+    "eslint": "7.29.0",
     "eslint-plugin-import": "2.23.4",
     "eslint-plugin-node": "11.1.0",
-    "eslint-plugin-promise": "4.3.1",
-    "husky": "6.0.0",
+    "eslint-plugin-promise": "5.1.0",
```
mightyiam, line 81: Changelog isn't helpful in figuring out what the breaking change is. But we're not using it directly, buy via upstream, so OK I guess.

package.json
```diff
@@ -71,21 +71,21 @@
...
+    "husky": "7.0.0",
```
mightyiam, line 82: Breaking change is merely dropping unsupported Node.js version.

PR #802, opened 2022-04-21
fix: use typescript-eslint version of space-infix-ops rule

davidyuk:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

Replaced `space-infix-ops` with `@typescript-eslint/space-infix-ops` as proposed in https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/space-infix-ops.md

This fixes the spacing around `|` in type definitions, like `arg: string | number` (instead of `string|number`).

I'm not sure that this would work here, because I'm using these configs with eslint@8

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -65,6 +65,7 @@ test('export', (t): void => {
           quotes: 'off',
           semi: 'off',
           'space-before-function-paren': 'off',
+          'space-infix-ops': 'off',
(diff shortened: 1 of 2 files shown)


mightyiam:
Thank you, @davidyuk .

PR #1123, opened 2023-05-05
build: rm extraneous --ext option usage in lint script

mightyiam:


mightyiam:
No, that's not correct.

PR #1169, opened 2023-06-30
fix: rules are at top level (not under overrides)

### LINKED ISSUE #1088: Option to include JS-Files
**What version of this package are you using?**
34.0.1

**What problem do you want to solve?**
I'm working with a codebase consisting of .js, .svelte and .astro files. All of them including JSDoc annotated JavaScript code with full type-information. All of this is correctly parsed by @typescript-eslint/parser. If I enable the rules globally or by using `plugin:@typescript-eslint/recommended-requiring-type-checking` the rules like no-floating-promises work in all files without problem.

I would like to use the ts-standard rules in this codebase, but I see no way to enable the rules globally or for other file-types, since the shared config only enables the rules in the override-section for ts and tsx files.

**What do you think is the correct solution to this problem?**
Maybe provide an option to enable the rules globally instead of just inside the override for ts-files? Or is there a way I could reference the rules in my ESLint-Config like this: ```rules: { ...standardWithTypescript.getRules(), ...} ``` ?

**Are you willing to submit a pull request to implement this change?**
Yes

### LINKED ISSUE #1149: Our config to not be under the overrides property
#43 was never a good idea. We mean to revert it, probably as a breaking change.

This happens to be the highest priority issue at this time.

mightyiam:
BREAKING CHANGE: the rules are provided at the top level,
instead of under an `overrides` property.
Providing the rules under the `overrides` property was never a good idea.
It prevents specifying which files the rules apply to (e.g. `[*.js, *.ts]`).
I apologize.
To migrate, make sure that your `extends` property is under an [`overrides` item][overrides].
An example is in the readme.
To help verify your configuration,
you could obtain a list of files that will be linted, this way:
`DEBUG=eslint:cli-engine npx eslint <path>`.

[overrides]: https://eslint.org/docs/latest/use/configure/configuration-files#how-do-overrides-work

closes #1149
closes #1088

Co-authored-by: Rostislav Simonik <rostislav.simonik@technologystudio.sk>

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,11 @@
 {
-  "extends": "./lib/index.js",
-  "parserOptions": {
-    "project": "./tsconfig.json"
-  }
+  "overrides": [
+    {
+      "files": ["*.js", "*.ts"],
+      "extends": "./lib/index.js",
+      "parserOptions": {
+        "project": "./tsconfig.json"
+      }
+    }
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "compile": "tsc",
-    "lint": "eslint --ext .js,.ts .",
+    "lint": "eslint .",
     "editorconfig": "editorconfig-checker",
     "unit": "ava",
     "test": "run-s clean-artifacts editorconfig compile lint unit",
--- a/readme.md
+++ b/readme.md
@@ -49,20 +49,28 @@ npm install --save-dev \
 
 # Example config
 
-Here is an example `.eslintrc.js`:
+Here is an example `.eslintrc.js`.
+Pay close attention to the `files` property, because it [determines which files are linted][specifying-target-files-to-lint].
 
 ```js
 module.exports = {
-  extends: 'standard-with-typescript',
-  parserOptions: {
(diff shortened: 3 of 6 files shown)


LinusU:
Upgrading from `eslint-config-standard-with-typescript` 35 to 36 led to TypeScript linting errors being reported in JavaScript files:

```text
❯ eslint --ignore-path .gitignore .

/Users/linus/coding/loco/admin/scripts/codegen.js
  21:16  error  Missing return type on function  @typescript-eslint/explicit-function-return-type

/Users/linus/coding/loco/admin/scripts/configure.js
  25:16  error  Missing return type on function  @typescript-eslint/explicit-function-return-type
```

Unfortunately I don't have more time to investigate this right now, will stick to 35 in the mean time...

mightyiam:
Control this using `overrides`. See README for example, albeit simple one.

LinusU:
Before, as far as I understood it, it would lint .js files using the regular standard rules, and .ts file with the TypeScript rules. It seems like now TypeScript rules are being applied to .js files, which makes them impossible to resolve (i.e. the error above tells me to annotate the return type on the function, but that is only possible in TypeScript)

mightyiam:
Thank you for explaining. What would have prevented this in the previous version? I'll have to take a look. There's a session tomorrow.

PR #1260, opened 2023-09-11
feat(docs): replace npm@<7 with yarn

gpoole:

[x] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

Update the install command in the readme for peer dependencies to quote strings with characters interpreted by shells like bash and zsh.

The install command as currently given is not usable in zsh without escaping the `^`s.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -29,15 +29,15 @@ This can be achieved by:
 
 And both pin/lock to the same version.
 
-# npm@<7
+# Yarn
 
-`npm@<7` does not automatically install `peerDependencies`,
+Yarn does not automatically install `peerDependencies`,
 so if that's what you're using, install them manually.
 Here is an example, but use it only for reference,
 because your decisions regarding version ranges and range specifiers may vary.
 
 ```
-npm install --save-dev \
+yarn add --dev \
   typescript@\* \
   eslint@^8.0.1 \
   eslint-plugin-promise@^6.0.0 \
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -367,11 +367,11 @@ test('Exported rule values do not reference eslint-config-standard ones', (t) =>
   }
 })
 
-test('npm install args in readme satisfy peerDeps', async (t) => {
+test('yarn install args in readme satisfy peerDeps', async (t) => {
   const { pkgJson, pkgPath, ourPeerDeps } = await getPkgDetails()
   const readme = (await readFile(resolve(pkgPath, '..', 'readme.md'))).toString()
-  const match = readme.match(/```\n(npm install .*?)```/s)
-  if (match === null) throw new Error()
+  const match = readme.match(/```\n(yarn add .*?)```/s)
+  if (match === null) throw new Error("couldn't find yarn add")
   if (match.length === 0) throw new Error('failed to find code block')
   if (match.length > 2) throw new Error('matched multiple code blocks')
   const installArgRanges = Object.fromEntries(


mightyiam:
Does anyone still such old npm?

Wouldn't we rather remove this altogether?

gpoole:
Yeah not sure about that version of NPM. The reason I needed it was for Yarn since (unless I've missed something) it doesn't seem to install peer deps automatically either. How about keeping it and changing it to a Yarn command?

mightyiam:
Really? Yarn behaves this way? Well... soon some day we won't have plugins as peerDeps. So, if anything, I would remove this.

mightyiam:
Sorry, I didn't mean to close. Would you like to remove it?

gpoole:
Yes, looks like this is Yarn's chosen behaviour for the present anyway @mightyiam: https://github.com/yarnpkg/yarn/issues/1503#issuecomment-753624157

I do think it's useful for Yarn users right now even if you're planning to remove it later, so would you consider keeping it in Yarn form until the time peer deps aren't required at all? I think removing it entirely will make life harder for Yarn users, since we'd have to figure out the install command for the peer deps from scratch instead of just modifying the handy npm command you have there, so if you'd rather not switch it to a Yarn command I'd say I'd definitely vote for leaving it alone.

mightyiam:
Okay, so we should replace the existing section with a yarn specific one.

gpoole:
Updated with a Yarn sample, does that look ok?

readme.md
```diff
@@ -29,22 +29,22 @@ This can be achieved by:
...
+```sh
+yarn add \
   "typescript@*" \
   "eslint@^8.0.1" \
   "eslint-plugin-promise@^6.0.0" \
   "eslint-plugin-import@^2.25.2" \
   "eslint-plugin-n@^15.0.0" \
   "@typescript-eslint/eslint-plugin@^6.4.0" \
-  eslint-config-standard-with-typescript@latest
+  eslint-config-standard-with-typescript
```
mightyiam: Is this change related?

readme.md
```diff
@@ -29,22 +29,22 @@ This can be achieved by:
...
-`npm@<7` does not automatically install `peerDependencies`,
+Yarn does not automatically install `peerDependencies`,
 so if that's what you're using, install them manually.
 Here is an example, but use it only for reference,
 because your decisions regarding version ranges and range specifiers may vary.
 
-```
-npm install --save-dev \
+```sh
+yarn add \
```
mightyiam: Is there no equivalent to `--save-dev`?

readme.md
```diff
@@ -29,22 +29,22 @@ This can be achieved by:
...
 
-`npm@<7` does not automatically install `peerDependencies`,
+Yarn does not automatically install `peerDependencies`,
 so if that's what you're using, install them manually.
 Here is an example, but use it only for reference,
 because your decisions regarding version ranges and range specifiers may vary.
 
-```
-npm install --save-dev \
+```sh
```
mightyiam, line 39: Unrelated change.

readme.md
gpoole: There is, I've added it in a second commit after this one.

readme.md
gpoole: If you specify a version explicitly, `yarn add` will put the literal version specified into your `package.json`, unlike npm which will resolve `latest` to a specific patch version range and write that. Leaving the version specifier off will implicitly request `latest`.

readme.md
gpoole: Reverted

mightyiam:
Seems like a test needs to be amended.

PR #1462, opened 2024-03-16
feat: rename to eslint-config-love

mightyiam:


mightyiam:
While it doesn't show up in the search, it does exist. It's deprecated.
https://www.npmjs.com/package/eslint-config-love

Oops.

mightyiam:
The author of that package gifted it to us. Thank you, @DaBultz.

PR #2198, opened 2025-09-18
feat: add test-friendly overrides and withTests() helper; document us…

rickyzakariap:
…age in README


rickyzakariap:
thanks for the answer!

Issue #2197, opened 2025-09-18
Consider additional rules


mightyiam:
There's a list of rules to consider in the file [`src/test/_rules_to_consider.ts`](https://github.com/mightyiam/eslint-config-love/blob/main/src/test/_rules_to_consider.ts).

PR #2302, opened 2026-03-09
feat!: eslint v10 support

### LINKED ISSUE #2293: feat: support eslint@^10

slifty:
This PR adds support for ESLint 10.  In addition to version bumps, this updates the way we load the rules from the @eslint/js plugin, since the linter no longer provides a mechanism to access default rules.  One byproduct of this is that deprecated rules are no longer accessible, which I think is just the state of the API in v10.

Resolves #2293


package.json
```diff
@@ -95,7 +96,27 @@
     "semver": "7.7.4",
     "type-fest": "5.4.4",
     "typescript": "5.9.3",
-    "typescript-eslint_bottom": "npm:typescript-eslint@8.53.0"
+    "typescript-eslint_bottom": "npm:typescript-eslint@8.56.0"
+  },
+  "overrides": {
```
slifty: I suspect that this block means you would *not* want to ultimately accept this PR (these are all the dependencies that don't officially support eslint 10 yet).

That being said there are enough other changes in this PR needed to support estlint@10 that I wanted to open it for consideration in order to potentially iterate on the "bigger" questions that are opened by the various internal API changes.  Once the dependencies update I can bump the PR to use those versions and take out these overrides.

mightyiam:
Okay, I've made a bit of progress and next session is same time next week, so feel free to make progress before then.

mightyiam:
Since https://github.com/eslint-community/eslint-plugin-promise/issues/616 is done the last remaining blocker is https://github.com/import-js/eslint-plugin-import/issues/3227.

mightyiam:
Should we switch to eslint-plugin-import-x?

slifty:
I saw that lib in the thread -- I feel torn since plugin-import *is* maintained by some solid folks... that said, the long delays in supporting current eslint versions are real and hard to ignore.

vorobiovkirill:
@mightyiam  @slifty Any ETA?

mightyiam:
@vorobiovkirill do you have any input?

gtbuchanan:
I highly suggest moving to `eslint-plugin-import-x`. In addition to the issues brought up by the `eslint-plugin-import-x` README:
* `eslint-plugin-import` hasn't had new commits in almost a year
* [ESLint 9 EOL is 2026-08-06](https://eslint.org/version-support/)
* [XO/sindreshorus uses import-x already](https://github.com/xojs/xo)

slifty:
The v10 PR for eslint-plugin-import has some recent activity and looks *almost* at the finish line: https://github.com/import-js/eslint-plugin-import/pull/3230

Maybe a good approach could be to pick a deadline (e.g. if the import doesn't support v10 by July 14th, move to plugin-import-x)

darkbasic:
It has been merged!

slifty:
woohoo! I'll update the PR to reflect this

EDIT: It hasn't been released yet but [should be soon](https://github.com/import-js/eslint-plugin-import/issues/3227#issuecomment-4834815021)

takaki:
I think it's time to move to `eslint-plugin-import-x`.

ESLint 9 reached end of life today, and `eslint-plugin-import` still has no released version supporting v10 — the support is merged, but npm `latest` is 2.32.0 from June 2025. As long as this config depends on it, releases here stay gated on a release over there that no one can put a date on.

The cadence makes me uneasy about depending on it going forward. `import-x` supports v10 today.

mightyiam:
Agreed. PR welcome.

darkbasic:
@mightyiam give it a try, I let codex give this a shot: https://github.com/mightyiam/eslint-config-love/pull/2433

mightyiam:
#2434 has two primary costs:

1. plugin name change from `import` to `import-x` (why did a fork change that name?)
2. https://github.com/mightyiam/eslint-config-love/pull/2434#issuecomment-5216705402

Would it make more sense for this project to drop `eslint-plugin-import` without substitute in order to allow progress and possibly re-introduce it or something similar later?

slifty:
@mightyiam I'll add a commit to this PR that drops the import plugin entirely.  It's a bummer to go "backwards" from a ruleset PoV, but I do agree that it would be useful to decouple that migration from the upgrade to eslint 10.

If we merge this I (or whoever!) can make an issue to capture the desire to add import-plugin-esque functionality back

mightyiam:
Would you mind doing a PR that only removes eslint-plugin-import?

slifty:
I'll rebase on main / ditch that commit (sorry for my delay in opening a separate PR)

slifty:
@mightyiam I think this may be finally ready to bring over the edge / review!

src/test/compatibility.ts
```diff
@@ -1,5 +1,5 @@
 import test from 'ava'
-import type { TSESLint } from '@typescript-eslint/utils'
```
mightyiam, line 2: Do we by chance not need `@typescript-eslint/utils` any more?

src/types.d.ts
```diff
@@ -24,13 +24,12 @@ declare module 'eslint-plugin-promise' {
...
-  const plugin: TSESLint.FlatConfig.Plugin
+  import type { ESLint } from 'eslint'
+  const plugin: ESLint.Plugin
   export default plugin
 }
 declare module 'eslint_bottom' {
-  import { TSESLint } from '@typescript-eslint/utils'
-  export class ESLint extends TSESLint.FlatESLint {}
+  export { ESLint } from 'eslint'
 }
```
mightyiam: Is there a way to alias the entire package, intead?

package.json
```diff
@@ -56,8 +56,13 @@
     "test": "run-s clean-artifacts format/check compile unit lint",
     "unit": "ava"
   },
+  "overrides": {
+    "eslint-plugin-promise_bottom": {
+      "eslint": "$eslint"
+    }
+  },
```
mightyiam: What is this for, please?

slifty: Whoops that was a lingering issue from before / that package supports ESLint 10 now.  bumped to the latest version and removed this override.

src/test/compatibility.ts
slifty: I think it's used elsewhere but a refactor could remove the need for it.  Maybe a good followup PR?

src/types.d.ts
slifty: Actually... it looks like we can take this declaration out now that eslint ships with types!

mightyiam:
Looking forward to review

darkbasic:
Is there anything I can help with to speed things up?

src/test/_util.ts
```diff
@@ -60,9 +60,7 @@ const { rules: ourRules_ } = exported
 if (ourRules_ === undefined) throw new Error('we seem to be exporting no rules')
 export const ourRules = ourRules_
 
-export const equivalents = [
-  ...new TSESLint.Linter({ configType: 'eslintrc' }).getRules().keys(),
-].filter((name) => {
+export const equivalents = Object.keys(js.configs.all.rules).filter((name) => {
```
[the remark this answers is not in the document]
mightyiam: I would prefer using that docs package for testing, if you don't mind.

src/test/compatibility.ts
mightyiam: Would removing it increase the complexity of this PR?

mightyiam (commented):
Was I holding this up? Sorry.

src/test/compatibility.ts
slifty: Yeah -- I don't know by how much until trying it, but it's the kind of thing that *could* become a yak shave.

Then again maybe it would be fast.

My suggestion would be to decouple it from the v10 update

mightyiam: Okay.

Echron:
What is the status on this? Need any help to finish this one?

mightyiam:
Thank you for asking, @Echron. A precursor pull request that uses

https://github.com/eslint/eslint/blob/v9.39.5/docs/package.json

as a source of lints metadata in our tests would be awesome.

slifty:
Sorry for the delay on my end -- pushing the requested change.

I'm not entirely sure I understood the request, so I pushed it as a separate commit.  From talking to my friendly robot it sounds like the more canonical approach is what was originally implemented, and that this requires accessing those rules in an odd way.

I pushed the WIP commit below for your consideration; I can tweak and fixup the original commit as desired!

mightyiam:
Thanks. Would you mind submitting the change that obtains the core rules metadata from that unpublished docs package as a standalone PR?

slifty:
Sure thing -- and I'll take it out of here; do you think this particular PR needs any more changes in order for us to release the v10 support at this point?

mightyiam:
> do you think this particular PR needs any more changes in order for us to release the v10 support at this point?

I prefer waiting until it has been rebased against a `master` that has the test change merged, if you don't mind.

slifty:
@mightyiam I may be misunderstanding, but currently the test change is built on *top* of this PR / is built with Eslint@10 in mind, which means I can't cherry pick that commit as written.

I may not fully grok the desired change change since I had understood it as being needed because of the version bump.  Rather than blocking on that knowledge transfer, maybe it would be better for someone else to make the test change in that separate PR, then I can rebase?

mightyiam:
In favor of #2457.

Thank you for your work and patience.

slifty:
A bit bummed to have put the time in and not get a co-author line, but glad the feature is in.

Best of luck as you continue to develop this ruleset!

mightyiam:
I could have easily done that and was actually sorry I didn't right after I merged it. If it helps, some perspective:

- this PR suggests some unnecessary changes
- #2457 is 8 changed lines excluding the lockfile.

If you'd like your name on some commits, please go through some rules and either reject or add them (#2197).

slifty:
Totally understood (sorry if I sounded overly sour -- truly I appreciate your work on this!)

PR #2433, opened 2026-08-06
Support ESLint 10 with eslint-plugin-import-x

### LINKED ISSUE #2293: feat: support eslint@^10

darkbasic:
## Summary

Adds ESLint 10 support and replaces `eslint-plugin-import` with `eslint-plugin-import-x`.

This builds on the investigation in #2302 and removes the final dependency blocker without relying on npm overrides.

Closes #2293.

## Changes and decisions

- Raises the ESLint peer dependency to `^10.0.0`, tests against ESLint 10.8.0, and keeps a separate ESLint 10.0.0 installation to verify the lower bound.
- Replaces the `eslint-plugin-import` dependency and the exported `import/*` rule namespace with `eslint-plugin-import-x` and `import-x/*`. This is intentionally a breaking change rather than exposing the new plugin under the old package's name.
- Uses `eslint-plugin-import-x` 4.17.0 as the minimum supported version. Version 4.16.2 introduced an `@package-json/types` declaration dependency that fails this repository's TypeScript 6 compilation. Version 4.17.0 replaces that dependency with inline types, making it the first usable lower bound for this project.
- Replaces `import/enforce-node-protocol-usage`, which is not available in `import-x`, with `n/prefer-node-protocol`. The explicit `>=16.0.0` option preserves enforcement for imports, exports, and `require()` calls instead of making the result depend on a consumer's `engines.node` field.
- Raises the minimum versions of `@eslint-community/eslint-plugin-eslint-comments` and `eslint-plugin-promise` to releases that declare ESLint 10 support. Their matching `_bottom` aliases were updated as well, so the compatibility test exercises real supported combinations without peer-dependency overrides.
- Replaces the removed legacy `Linter#getRules()` path with the public `@eslint/js` `all` config when enumerating core rules. Core rule accounting now covers current, non-deprecated rules; plugin deprecations are still read from rule metadata.
- Removes the deprecated `"always"` option from the `radix` rule while preserving its behavior under ESLint 10.
- Narrows the internal flat-config plugin type to the fields ESLint consumes. This avoids a type conflict caused by plugins that also bundle legacy configurations without weakening the exported config type.
- Updates the rule inventory for the `import-x` namespace and records its additional `no-rename-default` and `prefer-namespace-import` rules as rules to consider.

## Breaking changes

- ESLint 10 is now required.
- Consumers overriding or disabling `import/*` rules must change those rule IDs to `import-x/*`.
- The exported plugin key is now `import-x` instead of `import`.

## Verification

- `npm ci`
- `npm test`
  - formatting
  - TypeScript compilation
  - 18 unit and integration tests, including the node-protocol replacement
  - linting
- `npm audit --omit=dev` reports no production vulnerabilities.

## Notes
Made with codex and GPT-5.6 Sol xhigh


mightyiam:
Thank you. But... for me, this is a lot in one commit. Might be unavoidable, though. Can the plugin be replaced as a precursor PR?

darkbasic:
Sure, let's split this into two PRs.

darkbasic:
@mightyiam part1: https://github.com/mightyiam/eslint-config-love/pull/2434 part 2: https://github.com/mightyiam/eslint-config-love/pull/2435

PR #193, opened 2019-12-10
feat: upgrade plugin to v2.8.0

mightyiam:
BREAKING CHANGE: add rule @typescript-eslint/restrict-template-expressions
https://github.com/typescript-eslint/typescript-eslint/blob/v2.8.0/packages/eslint-plugin/docs/rules/restrict-template-expressions.md

BREAKING CHANGE: add rule @typescript-eslint/space-before-function-paren
https://github.com/typescript-eslint/typescript-eslint/blob/v2.8.0/packages/eslint-plugin/docs/rules/space-before-function-paren.md

BREAKING CHANGE: add rule @typescript-eslint/no-dynamic-delete
https://github.com/typescript-eslint/typescript-eslint/blob/v2.8.0/packages/eslint-plugin/docs/rules/no-dynamic-delete.md

Closes #175.

PR #2434, opened 2026-08-06
Replace eslint-plugin-import with import-x

darkbasic:
## Summary

Replaces `eslint-plugin-import` with `eslint-plugin-import-x` while retaining ESLint 9 support.

This is the precursor requested in #2302. Keeping the plugin migration separate leaves the ESLint 10 follow-up substantially smaller and easier to review.

## Changes and decisions

- Replaces the dependency and its minimum-version compatibility alias with `eslint-plugin-import-x` 4.17.0.
- Changes the exported plugin key and rule namespace from `import` and `import/*` to `import-x` and `import-x/*`. The new plugin is exposed under its actual name rather than hidden behind the old package's namespace.
- Uses 4.17.0 as the minimum because 4.16.2 introduces an `@package-json/types` declaration dependency that fails this repository's TypeScript 6 compilation. Version 4.17.0 replaces that dependency with inline types.
- Replaces `import/enforce-node-protocol-usage`, which is not available in `import-x`, with `n/prefer-node-protocol`. The explicit `>=16.0.0` option matches `eslint-plugin-n`'s fallback range and keeps enforcement from being weakened by consumer engine metadata or plugin settings. The replacement also covers `export *` and `process.getBuiltinModule()`.
- Adds a focused test for the replacement rule across static imports, named exports, dynamic imports, and `require()` calls, including its autofix.
- Updates the rule inventory for the new namespace and records the additional `no-rename-default` and `prefer-namespace-import` rules as rules to consider.
- Keeps two prerequisite type corrections in separate preparatory commits: the exported plugin map omits legacy configurations, and the rule inventory is typed as a rules map rather than a plugin object.

## Dependency tradeoff

`eslint-plugin-import-x` loads `unrs-resolver` through its built-in resolver. This adds platform-specific native or WASI packages and a `postinstall` fallback for missing optional bindings. There is currently no supported way to use import-x without it.

## Breaking changes

- Consumers overriding or disabling `import/*` rules must change those rule IDs to `import-x/*`.
- The exported plugin key is now `import-x` instead of `import`.
- The `import/node-version` setting no longer applies. The replacement rule uses its explicit version option and `eslint-plugin-n`'s builtin catalog.

## Verification

- `npm ci`
- `npm test`
  - formatting
  - TypeScript compilation
  - 18 unit and integration tests
  - linting
- `npm audit --omit=dev` reports one high-severity advisory for the locked `js-yaml` 4.3.0; the same transitive version is present on `master`, so it is not introduced by this change.


src/plugin-usage/import.ts
```diff
@@ -1,11 +1,10 @@
-import plugin from 'eslint-plugin-import'
+import { importX } from 'eslint-plugin-import-x'
```
mightyiam: Can this hunk be avoided?

src/test/expected-exported-value/_index.ts
```diff
@@ -1,20 +1,20 @@
 import type { TSESLint } from '@typescript-eslint/utils'
 import { parser, plugin as tseslintPlugin } from 'typescript-eslint'
 import eslintCommentsPlugin from '@eslint-community/eslint-plugin-eslint-comments'
-import importPlugin from 'eslint-plugin-import'
+import { importX } from 'eslint-plugin-import-x'
```
mightyiam: Should the new variable be `importXPlugin`?

src/plugin-usage/n.ts
```diff
@@ -11,6 +11,7 @@ const usage: PluginUsage = {
     'no-exports-assign': ['error'],
     'no-new-require': ['error'],
     'no-path-concat': ['error'],
+    'prefer-node-protocol': ['error', { version: '>=16.0.0' }],
```
mightyiam, line 14: Are you sure that this version value preserves existing behavior?

src/test/rules.ts
```diff
@@ -2,7 +2,7 @@ import test from 'ava'
 import { plugin as pluginTseslint } from 'typescript-eslint'
 import pluginEslintComments from '@eslint-community/eslint-plugin-eslint-comments'
 import pluginN from 'eslint-plugin-n'
-import pluginImport from 'eslint-plugin-import'
+import { importX } from 'eslint-plugin-import-x'
```
mightyiam: Should the new binding be `pluginImportX`?

src/plugin-usage.ts
```diff
@@ -5,11 +5,11 @@ import eslintRules from './plugin-usage/eslint.js'
 import nRules from './plugin-usage/n.js'
 import promiseRules from './plugin-usage/promise.js'
 import typescriptEslint from './plugin-usage/typescript-eslint.js'
-import importRules from './plugin-usage/import.js'
+import importXRules from './plugin-usage/import.js'
 
 export interface PluginUsage {
   pluginName: string
-  plugin: TSESLint.FlatConfig.Plugin | 'eslint'
+  plugin: TSESLint.FlatConfig.Plugins[string] | 'eslint'
```
mightyiam, line 12: Can this change be suggested atomically, please?

package.json
```diff
@@ -79,7 +79,7 @@
     "dedent": "1.7.2",
     "eslint": "9.39.5",
     "eslint_bottom": "npm:eslint@9.35.0",
-    "eslint-plugin-import_bottom": "npm:eslint-plugin-import@2.32.0",
+    "eslint-plugin-import-x_bottom": "npm:eslint-plugin-import-x@4.17.0",
```
mightyiam, line 82: Was the bottom version determined to be this exact one?

mightyiam (commented):
Should the `src/plugin-usage/import.ts` and `src/test/expected-exported-value/_import.ts` files be renamed?

Should the commit message mention the full package name?

Should the commit message include a reference to the relevant issue?

src/plugin-usage/import.ts
darkbasic: The import-form change can be avoided. I switched this back to the default export, so the line now changes only the package specifier.

src/test/expected-exported-value/_index.ts
darkbasic: Yes, changed to importXPlugin.

src/plugin-usage/n.ts
darkbasic: I rechecked both rule implementations. The explicit >=16.0.0 preserves the config’s default enforcement for the syntax supported by both rules and prevents a consumer’s engines.node, settings.n, or settings.node from disabling it.
It is also eslint-plugin-n’s fallback and the lowest simple continuous range supported for both ESM and require(); the rule additionally supports the Node 14.18 line.
The rules are not perfectly equivalent. The old rule derives its version from the Node runtime or import/node-version, while the replacement uses its own builtin catalog and additionally covers export * and process.getBuiltinModule(). I expanded the test to cover the shared static-import, named-re-export, dynamic-import, and require() behavior, including the rule option’s precedence over plugin settings.

src/test/rules.ts
darkbasic: Yes, changed to pluginImportX.

src/plugin-usage.ts
darkbasic: Done. This is now the separate 9a2fc10 commit, containing only the two flat-config type changes.

package.json
darkbasic: Yes. I tested 4.16.2, the first release with ESLint 10 support, but it fails this repository’s TypeScript 6 compilation because of declarations from @package-json/types. Version 4.17.0 replaces that dependency with inline types and is the first version that compiles here. The lower-bound compatibility test also passes with exactly 4.17.0.

darkbasic:
> Should the src/plugin-usage/import.ts and src/test/expected-exported-value/_import.ts files be renamed?

Yes. I renamed them to src/plugin-usage/import-x.ts and src/test/expected-exported-value/_import-x.ts, and updated all references.

> Should the commit message mention the full package name?

Yes. The subject is now feat!: replace eslint-plugin-import with eslint-plugin-import-x

> Should the commit message include a reference to the relevant issue?

Yes. I added Issue #2293 feat: support eslint@^10 to the commit message.

src/test/rules.ts
```diff
@@ -23,15 +23,16 @@ const knownEslintRules = new TSESLint.Linter({
...
 if (pluginEslintComments.rules === undefined) throw new Error()
 if (pluginN.rules === undefined) throw new Error()
-if (pluginImport.rules === undefined) throw new Error()
 if (pluginPromise.rules === undefined) throw new Error()
 if (pluginTseslint.rules === undefined) throw new Error()
 
-const rulesets: Array<[TSESLint.Linter.Plugin, string]> = [
+type PluginRules = NonNullable<TSESLint.FlatConfig.Plugins[string]['rules']>
+
+const rulesets: Array<[PluginRules, string]> = [
```
mightyiam, line 31: Can this change be made atomically, as well, please?

mightyiam (commented):
How do you feel about this eslint-plugin-import-x depending on unrs-resolver?

src/test/rules.ts
darkbasic: Done. I moved the rule-map type correction into its own commit, 81fe3e5, before the import-x migration. That commit only changes rulesets from a list of plugin objects to a list of rule maps, with the type derived from the flat-config plugin rules property. It passes the full test suite against the pre-migration dependency set.

darkbasic:
> How do you feel about this eslint-plugin-import-x depending on unrs-resolver?

I do not love it. After digging into the dependency chain, I think it is the clearest downside of this migration.
unrs-resolver is not optional in practice. eslint-plugin-import-x loads its built-in Node resolver when the plugin is imported, and that resolver imports unrs. Unrs then loads either a prebuilt native binding or its WASI fallback. Import-x cites package exports support, caching, and performance as the reasons for using it ([documentation](https://github.com/un-ts/eslint-plugin-import-x/blob/v4.17.0/README.md#differences)). The cost is a collection of platform-specific packages and a postinstall recovery path that can install or download a binding when optional-dependency installation fails.
The script is not required when the package manager installs the correct optional binding. npm blocked it in this checkout and the tests still passed. There are nevertheless open reports of native bindings failing in some pnpm and editor setups ([import-x #429](https://github.com/un-ts/eslint-plugin-import-x/issues/429)).
The supply-chain history is the more worrying part. The malicious napi-postinstall@0.3.1 release was published in July 2025 ([OSV MAL-2025-6025](https://osv.dev/vulnerability/MAL-2025-6025)), and the import-x and unrs ranges available at the time could resolve it. That version has since been removed from npm, and the current ^0.3.4 range could not select it in any case, so it is not an active installation risk. I still think the incident is relevant though, but I've lost count on how many packages have been compromised nowadays.
The graph locked here resolves unrs-resolver@1.12.2 and napi-postinstall@0.3.4 with integrity hashes. All three packages also list the same sole npm maintainer, so they are not independent maintainer boundaries.
I could not find a supported way to use import-x without unrs. There is an open issue to migrate import-x to oxc-resolver ([import-x #412](https://github.com/un-ts/eslint-plugin-import-x/issues/412)), and the unrs maintainer says the postinstall script will be removed in its next major release ([unrs #193](https://github.com/unrs/unrs-resolver/issues/193)). Neither change is available in the released import-x version. oxc-resolver also uses native bindings, so that migration would address the install script but not the native-code dependency.
My conclusion is that the dependency can be justified because supporting ESLint 10 matters more than avoiding native bindings and lifecycle scripts, but it is a real cost.

darkbasic:
@mightyiam what about making eslint-plugin-import-x an optional peer dependency? We could even support both import and import-x once the former releases a new version.

mightyiam:
Native deps are a big issue in this ecosystem. I suppose they have their place, but... installing them via npm might not be the right move. I wouldn't like that in this project.

PR #2435, opened 2026-08-06
eslint 10

### LINKED ISSUE #2293: feat: support eslint@^10

darkbasic:
## Summary

Adds ESLint 10 support on top of the `eslint-plugin-import-x` precursor PR.

This continues the work from #2302 now that every configured plugin has a released version compatible with ESLint 10.

Closes #2293.

## Changes and decisions

- Raises the ESLint peer dependency to `^10.0.0`, tests against ESLint 10.8.0, and keeps a separate ESLint 10.0.0 installation to verify the lower bound.
- Raises the minimum versions of `@eslint-community/eslint-plugin-eslint-comments` and `eslint-plugin-promise` to releases that declare ESLint 10 support. Their `_bottom` aliases are updated as well, so the lower-bound test uses supported combinations without npm overrides.
- Replaces the removed legacy `Linter#getRules()` path with the public `@eslint/js` `all` config when enumerating core rules. Core rule accounting covers current, non-deprecated rules; plugin deprecations are still read from rule metadata.
- Removes the deprecated `"always"` option from `radix` while preserving its behavior under ESLint 10.
- Updates the compatibility-test declarations and config type to the ESLint 10 APIs.

## Breaking changes

- ESLint 10 is now required.

## Verification

- `npm ci`
- `npm test`
  - formatting
  - TypeScript compilation
  - 18 unit and integration tests
  - linting
- `npm audit --omit=dev` reports no production vulnerabilities.


darkbasic:
I've amended last commit to preserve authorship since it was heavily inspired by https://github.com/mightyiam/eslint-config-love/pull/2302

